PR #70, opened 2015-03-28
cascadia paths

stennkool:
Go get cannot install goquery due to cascadia having moved from google code to github.
The source files have been changed to make use of the new cascadia sources.

### DIFF
--- a/expand.go
+++ b/expand.go
@@ -1,8 +1,8 @@
 package goquery
 
 import (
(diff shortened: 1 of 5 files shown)


cryptix:
Nice, thanks!

Friendly warning: I presume @PuerkitoBio will ask you to run gofmt on the code. :smile:

PuerkitoBio:
Thanks, fixed with #68

PR #73, opened 2015-04-16
Selection.AttrOr shorthand method

piotrkowalczuk:


PuerkitoBio:
Hi Piotr,

My first reaction was "no, I won't bloat the API one helper at a time" but then I checked the doc and this is the only "comma-ok"-style func where it would make sense to do this, and to be honest I would have liked such a helper a couple of times in the past. So all this to say: thanks for the PR, I will merge it.

Martin

PR #80, opened 2015-06-02
Add func (s *Selection) Val() string, for input[type=text|hidden|radio|checkbox], textarea, select

mantyr:
Example Doc4().Find("[name=name]").Val()

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,6 @@ The [BSD 3-Clause license][bsd], the same as the [Go language][golic]. Cascadia'
 [bsd]: http://opensource.org/licenses/BSD-3-Clause
 [golic]: http://golang.org/LICENSE
 [caslic]: https://github.com/andybalholm/cascadia/blob/master/LICENSE
-[forkfrom]: https://github.com/github.com/PuerkitoBio
 [doc]: http://godoc.org/github.com/PuerkitoBio/goquery
 [index]: http://api.jquery.com/index/
 [gonet]: https://github.com/golang/net/
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # goquery - a little like that j-thing, only in Go
+**forked from [PuerkitoBio/goquery][PuerkitoBio/goquery]**
 
 [![build status](https://secure.travis-ci.org/PuerkitoBio/goquery.png)](http://travis-ci.org/PuerkitoBio/goquery)
 
@@ -30,6 +31,7 @@ Please note that because of the net/html dependency, goquery requires Go1.1+.
 
 **Note that goquery's API is now stable, and will not break.**
 
+*    **2015-05-17** : Add `Val` helper method to return the value for input[type=hidden|text|radio|checkbox], textarea. [Oleg Shevelev][mantyr].
 *    **2015-04-20** : Add `AttrOr` helper method to return the attribute's value or a default value if absent. Thanks to [piotrkowalczuk][piotr].
 *    **2015-02-04** : Add more manipulation functions - Prepend* - thanks again to [Andrew Stone][thatguystone].
 *    **2014-11-28** : Add more manipulation functions - ReplaceWith*, Wrap* and Unwrap - thanks again to [Andrew Stone][thatguystone].
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Please note that because of the net/html dependency, goquery requires Go1.1+.
 
 **Note that goquery's API is now stable, and will not break.**
 
-*    **2015-05-17** : Add `Val` helper method to return the value for input[type=hidden|text|radio|checkbox], textarea. [Oleg Shevelev][mantyr].
+*    **2015-05-17** : Add `Val` helper method to return the value for input[type=hidden|text|radio|checkbox], textarea, select. [Oleg Shevelev][mantyr].
 *    **2015-04-20** : Add `AttrOr` helper method to return the attribute's value or a default value if absent. Thanks to [piotrkowalczuk][piotr].
 *    **2015-02-04** : Add more manipulation functions - Prepend* - thanks again to [Andrew Stone][thatguystone].
 *    **2014-11-28** : Add more manipulation functions - ReplaceWith*, Wrap* and Unwrap - thanks again to [Andrew Stone][thatguystone].
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # goquery - a little like that j-thing, only in Go
-**forked from [PuerkitoBio/goquery][PuerkitoBio/goquery]**
+**forked from [PuerkitoBio/goquery][forkfrom]**
 
 [![build status](https://secure.travis-ci.org/PuerkitoBio/goquery.png)](http://travis-ci.org/PuerkitoBio/goquery)
 
@@ -106,6 +106,7 @@ The [BSD 3-Clause license][bsd], the same as the [Go language][golic]. Cascadia'
 [bsd]: http://opensource.org/licenses/BSD-3-Clause
 [golic]: http://golang.org/LICENSE
 [caslic]: https://github.com/andybalholm/cascadia/blob/master/LICENSE
+[forkfrom]: https://github.com/github.com/PuerkitoBio
 [doc]: http://godoc.org/github.com/PuerkitoBio/goquery
 [index]: http://api.jquery.com/index/
 [gonet]: https://github.com/golang/net/
@@ -113,3 +114,4 @@ The [BSD 3-Clause license][bsd], the same as the [Go language][golic]. Cascadia'
 [wiki]: https://github.com/PuerkitoBio/goquery/wiki/Tips-and-tricks
 [thatguystone]: https://github.com/thatguystone
 [piotr]: https://github.com/piotrkowalczuk
+[mantyr]: https://github.com/mantyr
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
 # goquery - a little like that j-thing, only in Go
-**forked from [PuerkitoBio/goquery][forkfrom]**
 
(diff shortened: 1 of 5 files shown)


PuerkitoBio:
Hi Oleg,

Thanks for the PR, but I'm not sure about adding that, though... Since there's no live DOM, there's not much value to have Val(), it just returns whatever's in the value attribute (or checked, selected) so it can all be done relatively easily, without a hack, using normal selectors (like what you did to implement it). I can see it sometimes being used as helper in a project, but I don't feel it requires adding in the goquery package itself.

Martin

mantyr:
Hello. It was not easy to use goguery for quite simple and banal operation (see. for elegance jQuery) - receive the form values.

You're right, we can restrict the external function that do everything exactly the same as my proposed patch, but whether it will be convenient to me and other users? I think that is not so.

In the end, he who needs to this - find.

PuerkitoBio:
I'm afraid I don't understand the use case when you mention receive the form values? This parses a static html document, so to get e.g. the value of `txtUsername`, you do (untested, but should be along those lines):

```
username := doc.Find("[name=txtUserName]").AttrOr("value", "")
```

To find the checked option in a radio group:

```
opt := doc.Find("[name=radioName][checked]").AttrOr("value", "")
```

etc.?

mantyr:
Example: 
<form>
  <select name="name1">
    <option value="1">1</option>
    <option value="2" checked>2</option>
  </select>
  <input type="text" name="name1" value="test">
</form>
The question is, after submitting the form which value the server will be considered true? Answer
https://github.com/mantyr/goquery/blob/d6d2d23272087e75891132494d798f1c29b8d5c1/property.go#L18

PuerkitoBio:
It's the same thing, it can be done relatively easily and cleanly with a selector:

`v := doc.Find("[name=name1] option[checked]").AttrOr("value", "")`

Because there's no DOM, getting field values is just a matter of getting attribute values. I still don't understand the "after submitting the form" part, this has nothing to do with goquery, it deals with static html documents.

mantyr:
goquery in vacuum?:) ok, no problem, this open source.

offtop:
Rules of good product:
1. making the product quality
2. Think about how it will be used
3. do not make the banal and routine operations which can lead to errors in the final product
4. Make the world a better place

This conversation can be completed. Have a nice day.

mantyr:
Today:
```
v := doc.Find("[name=name1] option[checked]").AttrOr("value", "") // select
```
Tomorrow:
```
v := doc.Find("[name=name1]).Html()   // textarea
```
This code requires the support and advertence, because it can break down and an effective mechanism for checking "whether everything is correct," you do not offer.

PuerkitoBio:
Sorry that you take it like that, I was trying to better understand your use-case. jquery has val because it can change values dynamically - it has a live DOM - which is not the case in goquery. Otherwise getting those values is just getting any attribute. It has nothing to do with being in a vacuum, it's how it's implemented, based on the net/html parser.

mantyr:
Why do you think that the DOM not live in goguery? Sources of information that must be processed is constantly changing. In particular, it may be the logic that uses the developer working with existing html.
To me constantly meet:
1. polymorph forms
2. different layout-dependent:
- Time data acquisition
- Setting request (user-agent, country, ip mask, ipv4/ipv6, other...)
- The status of the account (admin, user, guest, other, pay functions)
- Authorized / not authorized
- User Settings
- Implementation of any external or internal conditions (the presence of certain data that change the entire layout for the visual presentation)

I have to check the operation of parsers for:
1. The lack of data
2. The correct lack of data
3. The correct handling errors when receiving data
4. Whether there was any authorization or application settings
5. Does the source of the data entered my settings (authentication, and other settings) for each request
6 signaling in different situations to adjust the parser (and the less I have to do it - the better)

mantyr:
```
v := doc.Find("[name=name1] option[checked]").AttrOr("value", "") // select
```
What if the form is not full? Not selected in the select and you need to take the default value?
Alternative solutions "not beautiful"
```
    if len(doc.Find("[name=name1]").Nodes) > 0 {
        s := doc.Find("[name=name1] option[checked]")
        v := ""
        if len(s.Nodes) == 0 {
            // default
            v = doc.Find("[name=name1] option").AttrOr("value", "") // first option
        } else {
            v = s.AttrOr("value", "")
        }
        fmt.Println(v)
    } else {
        fmt.Println("no select - ok, value need not")
    }
```
Too many code and so every time.

mantyr:
Example from real life. Set 1 very small settings for later use. The polymorphic form.
```
    headers := pparser.Headers{
        "REFERER": Url_page,
    }
    doc, _ := p.Page.GetQuery(Url_page, nil, headers) // set Cookies
    p.ViewState = doc.Find("[name=javax\\.faces\\.ViewState]").Val()

    // Polymorphic identifier in the form of options. First, we take a list of fields and then fill, remove the empty items and add additional fields
    options_name := make(map[string]string)

    doc.Find("#optionsForm").Find("input, textarea, select").Each(func(i int, s *goquery.Selection){
        options_name[s.AttrOr("name", "")] = ""
    })

    params := url.Values{}
    for key, _ := range options_name {
        if key != "" {
            key_select := strings.Replace(key, ":", "\\:", -1)
            key_select = strings.Replace(key_select, ".", "\\.", -1)
            val := doc.Find("[name="+key_select+"]").Val()
            if val != "" {
                params.Set(key, val)
            }
        }
    }
    params.Del("optionsForm:allchkBox")
    params.Set("optionsForm:wochkBox",         "on")
    params.Set("javax.faces.behavior.event",   "change")
    params.Set("org.richfaces.ajax.component", "optionsForm:wochkBox")
    params.Set("rfExt",                        "null")
    params.Set("AJAX:EVENTS_COUNT",            "1")
    params.Set("javax.faces.partial.ajax",     "true")

    doc, _ = p.Page.PostQuery(Url_page, params, headers)
    p.Set_wo = strings.Contains(doc.Find("#resultListForm\\:resultPanel0").Text(), "Office(s):wo")
```
Polymorphic forms are not ideal, they often show the same ID and it works in 80%. In a large sample you will notice a fully not working setup, and will be surprised that this did not notice before.

PuerkitoBio:
First, thanks for the previous comment (the one about too much code), that did make me find an error in AttrOr, it should return the default value if there is no node in the selection, not the empty string. That being said, you don't need all that code, even if there is no selection made, the single line will work as expected (once the bug is fixed) and will return the default value.

I'll take a look at the last comment later, from what I understand you have to get the value of a form element without knowing what's the type (text, select, etc.) of the element. That makes things clearer for me, always good to have a real-life use case as example.

mantyr:
```
doc.Find("#optionsForm").Find("input, textarea, select").Each(func(i int, s *goquery.Selection){
        options_name[s.AttrOr("name", "")] = ""
    })

    params := url.Values{}
    for key, _ := range options_name {
        if key != "" {
            key_select := strings.Replace(key, ":", "\\:", -1)
            key_select = strings.Replace(key_select, ".", "\\.", -1)
            val := doc.Find("[name="+key_select+"]").Val()
            if val != "" {
                params.Set(key, val)
            }
        }
    }
```
This was done because elements duplicated. Here is an example POST request that makes browser. The order of elements is important, it depends on the result.
```
*key value*
optionsForm	optionsForm
optionsForm:j_idt42-value	optionsForm:j_idt105
queryLanguageRadio	EN
optionsForm:stemChoice	on
optionsForm:j_idt60	10
resultListRadio	EN
optionsForm:j_idt72	ANID
optionsForm:j_idt72	EN_AB
optionsForm:j_idt72	IC
optionsForm:j_idt72	IN
optionsForm:j_idt72	DP
optionsForm:j_idt72	PA
optionsForm:j_idt76	table
optionsForm:j_idt80	CTR
optionsForm:j_idt80	ICF_MV
optionsForm:j_idt80	PAF_M
optionsForm:j_idt80	INF_M
optionsForm:j_idt80	DP
optionsForm:itemsPerClust...	10
optionsForm:defaultSearch...	search.jsf
optionsForm:selectedTabSi...	fpTab
optionsForm:selectedInter...	en
optionsForm:skinsList	classic
optionsForm:helpIpcChoice	on
optionsForm:allchkBox	on
optionsForm:wochkBox	on
javax.faces.ViewState	-6385525334393940090:-1003183206323592460
javax.faces.source	optionsForm:wochkBox
javax.faces.partial.event	change
javax.faces.partial.execu...	optionsForm:wochkBox @component
javax.faces.partial.rende...	@component
javax.faces.behavior.even...	change
org.richfaces.ajax.compon...	optionsForm:wochkBox
rfExt	null
AJAX:EVENTS_COUNT	1
javax.faces.partial.ajax	true
```

PuerkitoBio:
just to clear that up, you're not using goquery to parse a request to a web server, right? Some sentences you used and some code examples make me worried abot that. You should use the net/http.Request object for that.

That being said, I think you've made a valid point that Val() could be useful when you want the value of some input element without having to care about the element's type. However, it should be done exactly like jquery does it (returns value of first element in the selection, nothing if the first element is not a form element, multiple values if it's a list,etc.). It's not that trivial nor obvious (it may return multiple values). I'll think about it a bit more, don't expect anything soon.

mantyr:
It is true, I use already configured http.client, method which returns goguery DOM
```
func (p *Page) GetQuery(url string, params url.Values, headers Headers) (doc *goquery.Document, err error)
func (p *Page) PostQuery(url string, params url.Values, headers Headers) (doc *goquery.Document, err error)
```
As for the comparison with jQuery - I was guided by the following:
```
val_in_jquery := doc.Find("[name=name1]").Fitst().Val() // that normally even in the browser, the loop requires, to get the right value.
val_in_browser := doc.Find("[name=name1]").Last().Val()
```
I just have the priority to full browser emulation.

This conversation was definitely helpful. There is something to think about.

PR #82, opened 2015-07-24
Add a proxy option, compatible with the previous usage

wweir:


PuerkitoBio:
Hi,

Thanks for the PR, but `NewDocument` is really just a little helper, I don't want to add more network-related code to goquery as that doesn't belong in there anyway and short of passing a custom `http.Client`, there will always be different requirements (and no I don't want to add that `NewDocumentFromHTTPClient` :). There is `NewDocumentFromResponse` for when `NewDocument` doesn't do what you want (so you can use whatever options you want to retrieve the response).

Martin

PR #98, opened 2016-01-13
Keep document's charset and use it on returning values of nodes

otiai10:
I respect this very useful library. It helps me always!

This is just a suggestion.
It failed to return valid string when parsing multibyte encoded HTML document (like EUC_JP, ShiftJIS, or other multibyte encodings). So I fixed `Document` struct and `Attr` method a bit.

I wrote it as a wrap function of `Attr`, however, root document still have encoding, it might be better to let users use `document.DecodeString` in application layer.

Feel free to close this pull-request or I'm so glad if you leave some comment.

Thank you!

### DIFF
--- a/property.go
+++ b/property.go
@@ -17,7 +17,8 @@ func (s *Selection) Attr(attrName string) (val string, exists bool) {
 	if len(s.Nodes) == 0 {
 		return
 	}
-	return getAttributeValue(attrName, s.Nodes[0])
+	val, exists = getAttributeValue(attrName, s.Nodes[0])
+	return s.document.DecodeString(val), exists
 }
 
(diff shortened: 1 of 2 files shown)


otiai10:
oops

PuerkitoBio:
Hello,

Thank you for the kind words and thanks for trying this out. However, I prefer to keep encoding handling outside the library, and make it handle only utf-8 (as go.net/html does), to avoid having to depend on the text and charset packages.

You can see an example of how to handle this use-case in the wiki: https://github.com/PuerkitoBio/goquery/wiki/Tips-and-tricks#handle-non-utf8-html-pages

Thanks again,
Martin

otiai10:
OK, I think it's reasonable way.
Thank you very much!

PR #111, opened 2016-06-08
Added methods NewDocumentFromString and NewDocumentFromBytes

ANPez:


PuerkitoBio:
Hello Antonio,

Thanks for caring enough to send this PR, I appreciate it, but I don't want to add additional helpers to create a `*Document` (if anything, I'd remove the existing ones to leave just `New(io.Reader) (*Document, error)` and `NewFromNode(*html.Node) *Document`, but I can't do that).

As you see from the implementation, it's a single line call to transform the string or slice of bytes to a reader - an inevitable allocation since `html.Parse` requires an `io.Reader` anyway (and side-note, you should use `bytes.NewReader` and `strings.NewReader` when all you need is an `io.Reader`).

Thanks again, and have fun with your projects!
Martin

PR #142, opened 2016-12-27
Optimizes the *Selection.Text method and reduces allocations.

radovskyb:
This commit greatly optimizes the `*Selection.Text` method and highly reduces it's amount of allocations per op by inlining the `getNodeText` function since it's only used by `*Selection.Text`. 

This prevents the creation of many additional `byte.Buffer`'s every time `getNodeText` is called recursively and instead only creates a single `byte.Buffer`.


PuerkitoBio:
Looks good, thanks!

PR #147, opened 2017-01-20
Text nodes are written twice in selection.Text()

quipo:
I think this is a regression of commit c0b805c7dc712d7397ff1c76ae1ac6211a910e50 (pull request #142): 
If a document has an  or <a> tag which embeds a text node, calling Text() on it will return the text node twice, as both the  or <a> tag and the inner data nodes are considered as independent nodes in the loop, so the textual content is written twice.

### DIFF
--- a/property.go
+++ b/property.go
@@ -65,7 +65,7 @@ func (s *Selection) Text() string {
 	// Slightly optimized vs calling Each: no single selection object created
 	var f func(*html.Node)
 	f = func(n *html.Node) {
-		if n.Type == html.TextNode {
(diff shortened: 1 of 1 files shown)


quipo:
whilst this patch is safe, the error is probably in a wrapping library, which attempts "cleaning" the HTML by stripping some tags, without actually replacing the tag node with its text content only (it leaves the node.FirstChild property set to the tag's node child)

https://github.com/advancedlogic/GoOse/blob/master/parser.go#L16

Issue #151, opened 2017-01-29
proposal: Struct Tags


andrewstuart:
So I've been thinking for quite a long time that it would be awesome to be able to create structs with declarative tags for how an html document should be decoded into the struct. Much like one would do with the xml package, but using the syntax popularized by "that j-thing." 😄

So, for example:
```go
type Page struct {
	Resources []Resource `goquery:"#resources .resource"`
}

type Resource struct {
	Name string `goquery:".name"`
}

func main() {
  var p Page
  var r io.Reader
  // Get an io.Reader somehow
  goquery.NewDecoder(r).Decode(&p)
}
```

Any thoughts? Would this be something I could create a PR for?

PuerkitoBio:
Hey Andrew,

Yeah I also thought about something like that a while ago, I think it would be interesting but I don't think it has to be in the goquery package, it could very well be another package that uses goquery to achieve this.

If you feel like giving it a shot and build that package, let me know and I'd be glad to link to it from the docs!

Thanks,
Martin

barthr:
Sounds very cool, I would love to contribute to such package. Let me know!

PuerkitoBio:
You may want to take a look at this repo:

https://github.com/cathalgarvey/sqrape

barthr:
Thanks for sharing. Was looking through the code, there is a lot of reflection magic

PuerkitoBio:
You have to use reflection to get an arbitrary struct's fields and tags, that's how it's done in the json and xml packages too. That said, I haven't looked at the code, that's not an endorsement of its quality :) just that it seems to be more or less what Andrew had in mind.

barthr:
Yes, ofcourse. It just makes it a little bit harder to understand!

andrewstuart:
I did actually throw together an MVP yesterday that's implemented in-tree. Shouldn't be too hard to split it out, but I'm guessing it would probably be maintained better if it was in the same repo.

Most of the logic is handled by goquery already, so I was surprised by how little I had to implement.

I'd still like to handle more than just `Selection.Text()`, (attrs and HTML for examplre) but what I have already is pretty effective, at least for inner text.

https://github.com/andrewstuart/goquery/tree/unmarshaling

https://github.com/PuerkitoBio/goquery/compare/master...andrewstuart:unmarshaling

Let me know what your thoughts are. 😄

PuerkitoBio:
Hey Andrew,

I have to say it looks really good! A couple nitpicks:

[...]

PR #152, opened 2017-02-01
Add implementation for unmarshaling html directly into annotated go structs

andrewstuart:
The general principals I've tried to follow for the intuition and implementation are found in the documentation of the `Unmarshal` function. I figure we can move it around if necessary, or just link to it from the main package documentation.

In general, the tag structure is built around:
- A css selector
- A comma-separated list of "value selectors" that determine what is used to create primitive values. These should be intuitively in-order with respect to the primitive types in a type spec.
  - e.g. ```type S struct { MyMap map[int]map[string]string `goquery:"#foo,[bar],[baz],html"` }``` should behave intuitively, specifying values from left to right. There should be an int-valued `bar=...` attribute, a string-valued `baz=...` attribute, and the terminal value would be the html contained underneath an element that matches `[baz]` as a selector.

Let me know what feedback you have; this has grown a bit since my original proposal #151, but I really tried to keep the same clarity and update my naming conventions to use clearer same names everywhere.

This was fun to write. Looking forward to _actually_ using it instead of just writing tests. 😄


andrewstuart:
Ha, and I've already found a bug. I'm gonna add the HN front page as another test target to hopefully flush out some more. Meantime, feel free to comment on style, implementation, whatever. 😄

PuerkitoBio:
Thanks Andrew! I'll give it a good look this weekend.

PuerkitoBio:
I took a first look at it, I like how you thought about non-obvious cases like an embedded struct that has its own goquery tag. I think the behaviour you describe in the `Unmarshal` comment is pretty good, though there's nothing like real-life usage to figure out what works best, so at first we should document the new feature as experimental/beta and keep options open to break the API or behaviour as required.

There's one thing I'd like to change though, and that's the error handling. I understand that you tried to stay close to the json package and similar, but I'd rather follow the opaque error value as described here: https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully, so there's no exported error type in the API. There would be only `Unmarshaler`, `Unmarshal` and `UnmarshalSelection` added to the public API. (on that note, I'm on the fence if `Unmarshal` should take a `[]byte` or an `io.Reader` - I like how it matches the json/xml signatures, but I expect most calls will come from a file or http response's body).

In the simpler case (when you return a `*CannotUnmarshalError` with `Err` and `Reason` values), this is a simple matter of calling `errors.Wrap(err, "reason")` (no need for constants for the message, that may send the wrong message to users to assert the error based on the message string). That is, we'd add the `github.com/pkg/errors` as dependency and wrap the errors that way.

For more complex cases (when the error specifies the `Value` and `Field`/`Index`), I'd make them methods on a private error type, so that it can be asserted for behaviour as described in the blog post.

Does it make sense to you? I don't know if it's clear enough for you to take it from here, if you want to discuss it further, just let me know and we'll see how we go from here.

Thanks!
Martin

andrewstuart:
On `Unmarshal([]byte)` vs `Unmarshal(io.Reader)`, perhaps that's a good reason for us to consider following the typical pattern of having a `Decoder` type with a `Decode(interface{})` and `NewDecoder(io.Reader)`. I agree with what you said earlier about it potentially being misleading that goquery handles it as a stream, but the nice thing about the API is that it doesn't preclude the possibility that streaming might be a possibility someday.

As far as the error handling, I've watched Dave's talk and read his blog, and it's always seemed a bit redundant to me, since error is already opaque, being an interface already. I'd really rather encourage the use of built-in Go features (specifically type switching on errors over assertions) rather than getting a less-expressive variant from an external dependency. We'd also be forcing users to import and use Dave's package just to be able to programmatically inspect the errors they get, instead of just using Go the way it was designed.

I could go on for quite a while on why I discourage use of Dave's errors package, but the main driver in this case boils down to a sensible error message. I wanted to design from the start so that the fully-nested location of the error could be returned (which `encoding/json` is *just now* adding), and this requires a recursive type of some sort. It made most sense to me to just make that part of the error type, since we essentially need to unwind the stack, including loop positions, to give proper field names and array/slice indexes.

If you really want the error to be totally opaque, I'd be more open to not exporting the error type, but I personally don't agree with Dave that it's "brittle" just because we're exposing more information. That's specifically why I write methods that just return the `error` type rather than the concrete implementation: Go's `error` type is non-brittle by design. To change it, we can always just implement `error` with a different type and return that instead. In the worst case, a few people who made bad decisions about unchecked assertions might learn a valuable lesson. :stuck_out_tongue:

</soapbox>

Anyway, I'm definitely open to discussion on it, and in the end it's your package so it's 100% your decision. :smile:

PuerkitoBio:
Yeah I don't like adding `Decoder`-style API, it adds API surface without adding any value. For `Unmarshal`, thinking about the package user point of view, in most cases the user will have an `io.Reader` so if we stick with bytes, it will go like this:

```
reader := // get the reader somehow, usually via http.Response.Body or a file
bs, err := ioutil.ReadAll(reader)
// handle err
var s SomeStruct
if err := goquery.Unmarshal(bs, &s); err != nil {
  // handle err
}
```

and in Unmarshal, first thing we do: `bytes.NewReader(bs)`, because goquery only has a reader-based constructor, because `net/html` only has a reader-based Parse method. So I think it makes little sense to expose this function with a slice of bytes, it should be `Unmarshal(io.Reader, interface{}) error`.

I'll start a separate comment to address the errors.

andrewstuart:
Also, I meant to add I'm totally on board with making the Reason unexported and keeping the constants unexported as well.

PuerkitoBio:
About the errors, you mention that it boils down to sensible error messages, which is unrelated to whether or not we export a custom type - the `error` interface makes sure we can provide a sensible error message.

> If you really want the error to be totally opaque, I'd be more open to not exporting the error type

That's what I meant in my previous comment, sorry if that wasn't clear. It's not to provide less information, it's to provide as much, but differently. And I agree that we shouldn't force the `pkg/errors` dependency, that was a bad call, we could just implement the `Cause() error` method but even that is not needed IMO. I think there are 2 kinds of errors you handled with the same `CannotUnmarshalError` type - preconditions failing before starting the unmarshal process (e.g. failed to parse HTML document, invalid destination interface), and unmarshal value/type errors.

For preconditions, returning the root error is fine - the caller may assert for behavour, it won't have any specific one, and the message tells the whole story (an html.Parse error, or "non-pointer value" or "destination is nil", a bare `errors.New()` string-based error).

The second kind is where it may be useful to have an additional behaviour method, although for most purposes the error message itself should be complete enough to understand what's going on (the path where it failed, and the value that failed to be assigned).

Instead of maintaining a chain of errors with the `reflect.Value` at each level like you did, you should create an "unmarshaler" unexported struct that maintains the current state (the path in a stack, pushing e.g. "field" and "[index]", etc. and then joining the stack with a dot to get the full path when needed in an error), that is created to process each call to `Unmarshal*`, and have the unexported `unmarshalXxx` methods on it. So instead of creating an "unwinding stack" of errors, you create a simple (unexported type) error that immediately has the complete path to the target field that failed, and you just return that error to the top without wrapping it in other errors.

I think it would be useful to also capture the value that was attempted to be set, when the error is due to that value (i.e. in `unmarshalLiteral`, when ParseXxx fails). And the `V` and `FldOrIdx` fields on the error type are now unnecessary (since the decode state is maintained outside the error).

The hard part is naming the method on the returned error that the caller can assert on to check if this was the second category of error :) Maybe `Path() string` to get the path where it failed, and `Value() string` to get the value that failed to be assigned. I don't think most users will even care about extracting this information in the code, since the message will say it all (e.g. something like `"strconv.ParseInt: parsing 'abc': invalid syntax at root.field.[key]"`) and I can't think of a useful thing the program could do automatically with it, but who knows. We could keep them in the internal error type and not provide methods to get them for now, and see if it becomes needed eventually.

andrewstuart:
I believe what you describe (for the process of tracking error state) is essentially what I already have done, except that I haven't bothered maintaining state myself. Only in the case of an error do I bother building up the current state of the unmarshal process, which is already maintained by the stack since it's all done recursively. If an error occurs, we just make sure that we pass sufficient context back to the caller.

This is also done the way I've done it (as a linked list of errors) because this will be basically constant time when an error occurs, and linear time (to traverse the error linked list) _only_ if the caller calls `.Error()` for the string value of the error. So no computation or memory is wasted unless they want the full error message.

All in all, the "wrapped error/linked list" approach ends up being much more efficient and much less error prone than if I was to manually maintain some state in an unexported marshaler type, having to push/pop a self-maintained stack every time a recursion occurs. The go stack is going to be much more efficient without me getting in the way, anyway. I think this approach is much more important for packages that aren't using some underlying purpose-built parser, since they _have_ to maintain a state machine to properly parse text input into structured data. In our case, we have goquery (via `net/html`) already doing the parsing/decoding for us and we're just consuming the resulting tree.

As far as the error, I've now added a value attribute to the error message so that we can be a little more helpful when we attempt to unmarshal a string value that fails, at least in the case of `unmarshalLiteral`. I've also unexported the `V reflect.Value` field, since as you mentioned that's only needed internally to help build up the path string in a natural way (i.e. as fully valid Go syntax for the given type).

Aside from my general design reasons for just using recursion, TBH I'm not sure I'd be quite as excited to completely reimplement the error handling and parsing, given that we essentially have all the features already available.

While I still think there's much more utility in actually exporting a `CannotUnmarshalError` concrete type for direct inspection of relevant data (since errors semantically have only state and not behavior), I've kept that unexported for now. I'll probably raise the discussion again at some point and link back to this issue, but I'd rather get _something_ out there rather than holding up progress for design reasons. 😄 

Let me know if there's anything else missing in my updates (and if I'm missing something to get Travis to not barf on the `assert` testing dependency I added). 😄

PuerkitoBio:
Hey Andrew,

I was thinking about it while going through your PR, and you may hate my guts for it, but I think my initial hunch was correct - this should be in a separate package. It doesn't rely on any unexported feature of goquery, so there's technically nothing that prevents it from being done on the outside. This also sets a precedent for further requests to expand it even more (e.g. why this feature and not this other which also builds on goquery to provide something more). There's also the matter of integrating this implementation vs some other one, and maintaining more features. In the end, I feel there's just no good reason to have this in the same package.

I think there's already plenty of potential for expansion in goquery's core feature set (roughly, a jquery for Go) without adding additional (non-core, if you will) features. As a matter of fact I just merged such new core-goquery PR today.

I'm really sorry I made you spend time working on it on the premise that it would get merged, though I hope you feel like some of it was worth it nevertheless. As mentioned in the related issue, I will try to add a section in the readme to link to those kinds of packages, that build features on top of goquery.

Thanks again for your work on this, sorry about the change of mind but I do hope you end up releasing it.

Martin

andrewstuart:
Yeah, no worries at all, and definitely no hate 😄. I'm pretty sure I'll end up releasing it and let you know when I do with a README PR or something like that. As you mentioned, there's no real dependencies on internal APIs, so it'll be quite trivial to split out into a separate package. And even if it gets no traction, it was definitely worth learning some of the usage of go reflection, etc. :-)

PR #156, opened 2017-02-08
implement SetHtml and SetText

### LINKED ISSUE #123: How to replace text within some tag
Hi,
I want to postprocess static HTML (created by me).
Let's say that we have `<h1 class="message">placeholder</h1>` as part of document, and I want to replace `placeholder`, my selector is `.message`.

Currently I see only one option: use  `ReplaceWithHTML`, but it needs to:
- know, that used tag is _h1_;
- or figure it out via `NodeName` function - it seems unnecessarily complicated to construct replacement string in that way.
Both options needs to copy also all atributes.

I hope that I missed cleaner method, could you point it?

glebtv:
fixes #123

### DIFF
--- a/manipulation.go
+++ b/manipulation.go
@@ -270,6 +270,29 @@ func (s *Selection) ReplaceWithNodes(ns ...*html.Node) *Selection {
 	return s.Remove()
 }
 
+// Set nodes to specified nodes.
+func SetNodes(s *Selection, ns ...*html.Node) *Selection {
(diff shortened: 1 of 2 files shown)


PuerkitoBio:
Thanks! Will try to take a look this weekend.

PuerkitoBio:
Looks good, thanks! I unexported `SetNodes` (at least for now) and added a couple tests.

PR #157, opened 2017-02-11
Proposal for Text(): make <br> as newlines

hhh0pE:
It's useful for my project, maybe it will be useful for other users of this lib.

### DIFF
--- a/property.go
+++ b/property.go
@@ -65,6 +65,10 @@ func (s *Selection) Text() string {
 	// Slightly optimized vs calling Each: no single selection object created
 	var f func(*html.Node)
 	f = func(n *html.Node) {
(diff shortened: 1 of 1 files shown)


PuerkitoBio:
Hello,

I can see how it makes sense to want that behaviour, but I try as much as possible to keep things consistent with jquery behaviour-wise, and `text()` really just returns the text content of elements, with no special processing regarding tags. If you start inserting newlines for `<br>`, then what should you do for `<p>` and, to some extent, for any block-style element. Also, that would be a breaking change in behaviour.

For those reasons, I'm gonna pass on this PR. Thanks for your interest, though, and have fun using the package!
Martin

PR #175, opened 2017-04-08
Standardize HTML charset to UTF-8

dpcat237:
With this changes, I fixed the problem when charset of HTML isn't UTF-8.
Let me know if I should improve something or if would be necessary tests for new code.

### DIFF
--- a/type.go
+++ b/type.go
@@ -1,8 +1,10 @@
 package goquery
 
 import (
+	"bytes"
 	"errors"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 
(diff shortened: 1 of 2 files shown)


mna:
Hello Denys,

Thanks for your interest in goquery, but this change would add two 3rd party dependencies to goquery, with one that uses cgo, for all users of the package even if they don't need the conversion. I prefer to leave goquery "utf8-only" and have those that need it do the conversion before passing the reader to goquery, using the packages they prefer (because the official x/text/encoding package also supports this transformation, without relying on cgo).

Thanks, sorry for rejecting the PR but hopefully you understand the motivation,
Martin

dpcat237:
Hello Martin,

Thank you for the explanation. Yes, it makes sense. Okay, I'll do it passing the reader already with UTF-8 charset.

PS. Thanks for creating this library!

Regards,
Denys

PR #180, opened 2017-05-04
add Iteration method and tests

shellus:
Usage example:

```
	sel := Doc().Find(".hero-unit .row-fluid")

	for n := range  sel.Iteration(){
		t.Logf("node %v", n.Nodes[0].Data)
	}
```

### DIFF
--- a/iteration.go
+++ b/iteration.go
@@ -12,6 +12,22 @@ func (s *Selection) Each(f func(int, *Selection)) *Selection {
 	return s
 }
 
+// Each iterates over a Selection object, executing a function for each
+// matched element. It returns the current Selection object. The function
+// f is called for each element in the selection with the index of the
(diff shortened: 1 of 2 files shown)


mna:
Hello,

Thanks for your interest in goquery! Unfortunately, I'm gonna have to reject the PR for a couple reasons:

- It is not a jQuery method. Goquery tries to stay very close to the jQuery API as a design goal, with the special naming pattern described in the README.
- There is already (more than!) one simple way to iterate over the nodes in a selection, as mentioned in the wiki.

In addition to the example in the wiki, you could also do something like that to iterate:

```
	for i := 0; i < s.Length(); i++ {
		ss := s.Eq(i)
		fmt.Println(ss.Html())
	}
```

So I will not add a dedicated method for this, sorry.

Martin

Issue #173, opened 2017-04-04
Deprecate NewDocument and NewDocumentFromResponse


mna:
The documentation should mention that `NewDocument(url)` and `NewDocumentFromResponse(httpResponse)` are just helpers for quick prototyping/testing, but for "production" code, the HTTP handling should be done outside goquery, and call `NewDocumentFromReader(io.Reader)` when needed.

Many recent issues are raised because people expect a wide array of behaviour from those 2 functions which shouldn't have been added in the first place. Maybe they should be properly deprecated (as explained here: https://rakyll.org/deprecated/).

At the very least the docs should mention that "advanced" usage of those functions is not recommended. WIth better wording of course.

kakera87:
I am sorry I don'n use goQuery. I just newb TT

PR #203, opened 2017-08-28
Add second argument for POST request at NewDocument()

watanabeyu:
I want to get POST page.  
But this package is can't get it, so I made this PR.  
  
Usage example:
```
// POST page
doc, err := goquery.NewDocument("http://xxxxxxxxx", map[string]string{
  "url": "hoge",
  "foo": "1",
  "bar": "bar",
})

// GET page
// If want to GET page, not necessary second argument.
// same as until now
doc, err := goquery.NewDocument("http://xxxxxxxxx")
```

### DIFF
--- a/type.go
+++ b/type.go
@@ -31,12 +31,27 @@ func NewDocumentFromNode(root *html.Node) *Document {
 // NewDocument is a Document constructor that takes a string URL as argument.
 // It loads the specified document, parses it, and stores the root Document
 // node, ready to be manipulated.
-func NewDocument(url string) (*Document, error) {
+func NewDocument(u string, vr ...map[string]string) (*Document, error) {
 	// Load the URL
-	res, e := http.Get(url)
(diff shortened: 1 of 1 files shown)


mna:
Hello,

Unfortunately, the `NewDocument(url)` function should be deprecated (see #173 that I haven't "implemented" yet). In an ideal world, only `NewDocumentFromReader` would have been added to the package (well, maybe `...FromNode` too), as this is not the domain of goquery to deal with network requests.

So thanks for caring enough to send this PR, but for that reason I'm gonna have to reject it. To process the response from a POST or any other type of requests, the recommended approach is to use the stdlib (or the network requests package of your choice) to get the response, validate that the status code and any other characteristic of the response is "correct" (for whatever definition of "correct" works for your use-case) and then parse the body with goquery by calling `NewDocumentFromReader`.

Hope this helps,
Martin

PR #209, opened 2017-09-21
Helper method to read from string

sosedoff:
New helper method `NewDocumentFromString` would be helpful to create document that have sources other than http/io.reader. I found this be more convenient instead of doing `strings.NewReader(s)` all the time.

### DIFF
--- a/type.go
+++ b/type.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"strings"
 
 	"github.com/andybalholm/cascadia"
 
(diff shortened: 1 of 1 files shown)


mna:
Hello Dan,

Thanks for this, but as you've seen, it is merely a one-line helper and I don't feel it adds enough value to add to the package. In fact if I'm more interested in trimming down the number of `New...` functions in goquery - see #173 . And the underlying net/html package only accepts `io.Reader`s anyway, so it requires the creation of the reader regardless - there's no performance gain to be had i.e. to avoid some allocations.

For those reasons I will have to reject the PR, sorry.
Martin

sosedoff:
Not a big deal, as you've said its pretty much a one-liner, i'll just add that to my app instead.

PR #227, opened 2018-01-05
add NewDocumentFromBytes

rickypai:
Some interfaces like databases returns `[]byte` for data. This PR provides a shortcut to create a document from bytes.

### DIFF
--- a/type.go
+++ b/type.go
@@ -1,6 +1,7 @@
 package goquery
 
 import (
+	"bytes"
 	"errors"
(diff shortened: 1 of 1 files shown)


mna:
Hello Ricky,

Thanks for your interest in goquery, but as discussed in #209 , I don't want to add a `New...` function for bytes and/or strings, the existing `NewDocumentFromReader` should be used instead.

Thanks,
Martin

PR #231, opened 2018-01-17
Removed unnecessary conditional check

Strum355:
`IsMatcher()` returns false if `len(s.Nodes) !> 0`, so `Is()` was double checking if `len(s.Nodes) > 0`


mna:
Thanks!

PR #233, opened 2018-01-28
Fixes #232 - make .Slice's end argument optional.  If not provided, d…

### LINKED ISSUE #232: .Slice should have end be an optional parameter
In jQuery, the slice method only requires the first argument.

https://api.jquery.com/slice/
> The supplied start index identifies the position of one of the elements in the set; if end is omitted, all elements after this one will be included in the result.

The go code:

    func (s *Selection) Slice(start int, end ...int) *Selection

would work exactly as jQuery does - always use start, use end if it is provided as the second argument, and ignore any other arguments provided.

davidjwilkins:
make .Slice's end argument optional.  If not provided, default to the length of the nodes. Added unit test.

### DIFF
--- a/array.go
+++ b/array.go
@@ -36,14 +36,20 @@ func (s *Selection) Eq(index int) *Selection {
 
 // Slice reduces the set of matched elements to a subset specified by a range
 // of indices.
-func (s *Selection) Slice(start, end int) *Selection {
+func (s *Selection) Slice(start int, end ...int) *Selection {
+	var e int
+	if len(end) > 0 {
+		e = end[0]
+	} else {
+		e = len(s.Nodes)
+	}
 	if start < 0 {
 		start += len(s.Nodes)
 	}
-	if end < 0 {
-		end += len(s.Nodes)
+	if e < 0 {
+		e += len(s.Nodes)
 	}
-	return pushStack(s, s.Nodes[start:end])
+	return pushStack(s, s.Nodes[start:e])
 }
 
 // Get retrieves the underlying node at the specified index.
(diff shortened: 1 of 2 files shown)


mna:
Hello David,

We do try to keep goquery as close to jquery as possible, while keeping things type-safe and as Go-like as possible. In this case, this would be a breaking change, and it is kind of a hack to use the variadic arg approach to make a single parameter optional. I don't think this adds much value, especially for a breaking change, so I'm gonna pass on this.

Thanks for caring enough to send this, though, I appreciate it!
Martin

davidjwilkins:
I understand where you are coming from, but a few thoughts - first, I don't think this is a breaking change - any existing code continues to work just as before (if you tried to call .Slice with a single argument before, it wouldn't have compiled.  If you call it with two arguments, it works just as before).

As an example of the value that it adds - wehn dealing with a table where you don't need the first N rows, currently you have to do something like:

    rows := s.Find("tr")
    rows = s.Slice(3, rows.Length())
    rows.Find("td").Each(func(i int, s *goquery.Selection) { /* .... */)

On the other hand, with this, I could do:

    s.Find("tr").Slice(3).Find("td").Each(func(i int, s *goquery.Selection) { /* ... */ })

For me, not having to break out of the chain adds a lot of convenience.

mna:
Hello David,

Changing a function signature is always a breaking change, i.e. someone may store the `Slice` method into a variable or pass it around. Your use-case is interesting, though, thanks for raising that up. I'm still not a fan of the variadic `end ...int` signature to support this case, but I believe this deserves further thought. I'll reopen the related issue.

Thanks,
Martin

PR #235, opened 2018-01-29
Fixes #178 - Send context to parseHtml

### LINKED ISSUE #178: Append method ignores td th
If I try to `Append("<th>New</th>")`, the 'th' tag is always dropped. Same for td, tr

davidjwilkins:
This creates a new function `parseHtmlWithContext`, which behaves just like `parseHtml` but takes an `*html.Node` as context.  

It also create a function `eachNodeHtml` which will iterate over each node in a selection, check to see if we have parsed html for this node type already, and if not, will call `parseHtmlWithContext` with either the current node, or it's parent, depending on whether you passed `parent` as `true` or `false`.

It also changes `AfterHtml`, `AppendHtml`, `BeforeHtml`, `PrependHtml`, `ReplaceWithHtml`, `SetHtml`, `WrapHtml`, `WrapAllHtml`, and `WrapInnerHtml` to use one or both of these functions.

New tests were added for all except the `WrapHtml*` functions - changing these functions may have been unncessary - I couldn't think of a way to test this - Go automatically wraps `tr` elements in `tbody`, so I can't wrap those, and I can't think of any other situation that this would occur in.

### DIFF
--- a/manipulation.go
+++ b/manipulation.go
@@ -39,8 +39,14 @@ func (s *Selection) AfterSelection(sel *Selection) *Selection {
 // AfterHtml parses the html and inserts it after the set of matched elements.
 //
 // This follows the same rules as Selection.Append.
-func (s *Selection) AfterHtml(html string) *Selection {
-	return s.AfterNodes(parseHtml(html)...)
+func (s *Selection) AfterHtml(htmlStr string) *Selection {
+	return s.eachNodeHtml(htmlStr, true, func(node *html.Node, nodes []*html.Node) {
+		for _, n := range nodes {
+			if node.Parent != nil {
+				node.Parent.InsertBefore(n, node.NextSibling)
+			}
+		}
+	})
 }
 
 // AfterNodes inserts the nodes after each element in the set of matched elements.
@@ -85,8 +91,12 @@ func (s *Selection) AppendSelection(sel *Selection) *Selection {
 }
 
 // AppendHtml parses the html and appends it to the set of matched elements.
-func (s *Selection) AppendHtml(html string) *Selection {
-	return s.AppendNodes(parseHtml(html)...)
+func (s *Selection) AppendHtml(htmlStr string) *Selection {
+	return s.eachNodeHtml(htmlStr, false, func(node *html.Node, nodes []*html.Node) {
+		for _, n := range nodes {
+			node.AppendChild(n)
+		}
+	})
 }
 
 // AppendNodes appends the specified nodes to each node in the set of matched elements.
@@ -123,8 +133,14 @@ func (s *Selection) BeforeSelection(sel *Selection) *Selection {
 // BeforeHtml parses the html and inserts it before the set of matched elements.
 //
 // This follows the same rules as Selection.Append.
-func (s *Selection) BeforeHtml(html string) *Selection {
-	return s.BeforeNodes(parseHtml(html)...)
+func (s *Selection) BeforeHtml(htmlStr string) *Selection {
+	return s.eachNodeHtml(htmlStr, true, func(node *html.Node, nodes []*html.Node) {
+		for _, n := range nodes {
+			if node.Parent != nil {
+				node.Parent.InsertBefore(n, node)
+			}
+		}
+	})
 }
 
 // BeforeNodes inserts the nodes before each element in the set of matched elements.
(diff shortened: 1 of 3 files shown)


mna:
Hello David,

Thanks for this! I'll try to review it this weekend.

Martin

mna:
Sorry, couldn't quite make it last weekend. Should be good this week.

manipulation.go
```diff
@@ -572,3 +652,24 @@ func (s *Selection) manipulateNodes(ns []*html.Node, reverse bool,
 
 	return s
 }
+
+func (s *Selection) eachNodeHtml(htmlStr string, parent bool, fn func(n *html.Node, nodes []*html.Node)) *Selection {
```
mna: This would require a bit of documentation, explaining the purpose (in general, but also specifically of the nodesMap).

manipulation.go
```diff
@@ -572,3 +652,24 @@ func (s *Selection) manipulateNodes(ns []*html.Node, reverse bool,
...
+	nodesMap := make(map[html.NodeType][]*html.Node)
+	var context *html.Node
+	for _, n := range s.Nodes {
+		if parent {
+			context = n.Parent
+		} else {
+			context = n
+		}
+		if context != nil {
+			nodes, found := nodesMap[context.Type]
```
mna: The cache isn't quite right - it caches the parsed nodes based on the context node's type (TextNode, ElementNode, etc.) but the parser is more fine-grained than that - e.g. it won't parse a `"<tr><td>...</td></tr>"` the same way inside a `<table>` vs inside say a `<div>`. I think a better way would be to store it based on the equivalent of calling `goquery.NodeName` (except that this `NodeName` function should be refactored to call an unexported function that accepts an `*html.Node` so that it is faster when we already have a node).

Here's an example code that illustrates the issue (should probably be turned into a test case), in `d3`, the call to `AppendHtml` will add the `<tr>...</tr>` nodes both inside the `<table>` (which is fine) and the `<a>` (which isn't):

```
var insideA = `<html>
<body>
	<a class="x"></a>
</body>
</html>`

var insideTable = `<html>
<body>
	<table class="x"></table>
</body>
</html>`

var insideBoth = `<html>
<body>
	<table class="x"></table>
	<a class="x"></a>
</body>
</html>`

func main() {
	d1, err := goquery.NewDocumentFromReader(strings.NewReader(insideA))
	if err != nil {
		log.Fatal(err)
	}
	d2, err := goquery.NewDocumentFromReader(strings.NewReader(insideTable))
	if err != nil {
		log.Fatal(err)
	}
	d3, err := goquery.NewDocumentFromReader(strings.NewReader(insideBoth))
	if err != nil {
		log.Fatal(err)
	}

	s1 := d1.Find(".x").AppendHtml("<tr><td>Hello</td></tr>")
	s2 := d2.Find(".x").AppendHtml("<tr><td>Hello</td></tr>")
	s3 := d3.Find(".x").AppendHtml("<tr><td>Hello</td></tr>")

	o1, _ := goquery.OuterHtml(s1)
	o2, _ := goquery.OuterHtml(s2)

	fmt.Println(o1, "\n\n\n")
	fmt.Println(o2, "\n\n\n")

	s3.Each(func(i int, s *goquery.Selection) {
		o, _ := goquery.OuterHtml(s)
		fmt.Println(o, "\n\n\n")
	})
}
```

manipulation.go
```diff
@@ -272,8 +299,17 @@ func (s *Selection) ReplaceWithNodes(ns ...*html.Node) *Selection {
...
 // specified html string.
-func (s *Selection) SetHtml(html string) *Selection {
-	return setHtmlNodes(s, parseHtml(html)...)
+func (s *Selection) SetHtml(htmlStr string) *Selection {
+	for _, context := range s.Nodes {
+		for c := context.FirstChild; c != nil; c = context.FirstChild {
+			context.RemoveChild(c)
+		}
+	}
+	return s.eachNodeHtml(htmlStr, true, func(node *html.Node, nodes []*html.Node) {
```
mna: Here it passes `true` for `parent`, but I think it should be `false` (the context node is the actual node in the selection , `SetHtml` sets its html content).

manipulation_test.go
```diff
@@ -56,6 +56,13 @@ func TestAfterHtml(t *testing.T) {
 	printSel(t, doc.Selection)
 }
 
+func TestAfterHtmlContext(t *testing.T) {
```
mna: General comment for tests of methods that change the HTML, I think it's easier to read and setup the test by using a local string as data (instead of using the helper `DocXxx` functions), and parsing with `NewDocumentFromReader(strings.NewReader(data))`.

mna:
Superceded by #348 .

PR #239, opened 2018-03-07
Add a whitespace when encountering a br node

Freezerburn:
Prevents br nodes from causing text from after the br node to be
directly after the text before the br node without any whitespace.

### DIFF
--- a/property.go
+++ b/property.go
@@ -68,6 +68,8 @@ func (s *Selection) Text() string {
 		if n.Type == html.TextNode {
 			// Keep newlines and spaces, like jQuery
 			buf.WriteString(n.Data)
+		} else if n.Data == "br" {
+			buf.WriteString(" ")
 		}
 		if n.FirstChild != nil {
 			for c := n.FirstChild; c != nil; c = c.NextSibling {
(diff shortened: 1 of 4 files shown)


mna:
Hello Vincent,

First of all, thanks for caring enough about goquery to work on this PR, I appreciate it (kudos for adding a test too!). I understand the motivation, however there are a few concerns with the proposed solution. I don't like adding this specific special-case for `<br>` as this makes a precedent for many other cases: what about `<p>`? And `<hr>`, or some may even argue for `<div>` or any block element to have similar behaviour. And while you added a standard space here, some may want a newline, a non-breaking space, etc. There are many ways that this could be implemented and I don't feel like it makes sense for goquery to decide which way it is.

I think this could be part of a larger "pretty-printer" implementation that could work on a goquery selection (or more likely - as a more general solution -, on an `html.Node`, in which case it doesn't even have to be part of goquery nor to depend on it). I recall some discussion about this in a previous issue, that would be a pretty interesting project in itself.

And besides, the current behaviour is the same as jquery's, so there's a good reason to keep it this way too (without talking about how this could break stuff for existing code).

So for those reasons, I'm gonna have to decline this, sorry but I hope you understand the rationale!

Thanks,
Martin

PR #243, opened 2018-03-13
Added gnulnx/goperf to Related Projects

gnulnx:
First off thank you for such a fantastic library!  I had been using regex's to try and get src, and href attributes from HTML documents... At someone's request I looked at goquery and it was not only a perfect fit it was quite a bit faster than the initial regex I was using!  

In any event I saw that you were interested in related projects so I updated the README with my project.  If it's not related enough no biggy, but figured I'd submit any way.

gnulnx/goperf is a website performance testing tool that also fetches the static assets. (now using goquery). ;)

Thanks again.


mna:
Cool project! Thanks!

PR #276, opened 2019-01-02
Add 2 functions,TextWithoutJS\HtmlWithtoutJS. 

Dereking:
I need a function to get HTML\Text without script. Because I need plain text. I think those functions are need by many people.

### DIFF
--- a/property.go
+++ b/property.go
@@ -59,7 +59,7 @@ func (s *Selection) SetAttr(attrName, val string) *Selection {
 
 // Text gets the combined text contents of each element in the set of matched
 // elements, including their descendants.
-func (s *Selection) Text() string {
+func (s *Selection) Text(bIncludeJS bool) string {
 	var buf bytes.Buffer
 
 	// Slightly optimized vs calling Each: no single selection object created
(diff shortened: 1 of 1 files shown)


mna:
Hello Derek,

Thanks for your interest in goquery! I appreciate the PR but unless it is for a bug fix, it is best to discuss it in an issue first, to avoid some possible disappointment.

In this case, there is a simple enough way to do what you want without adding API surface:

```
var data = `
<html>
  <head>
    <title>Smoe html</title>
  </head>
  <body>
    <script src="file.js"></script>
    <h1>header</h1>
    <div>
      <p>A paragraph</p>
      <script>
        window.alert('hello');
      </script>
    </div>
  </body>
</html>
`

func main() {
	doc, err := goquery.NewDocumentFromReader(strings.NewReader(data))
	if err != nil {
		log.Fatal(err)
	}
	doc.Find("script").Remove()
	fmt.Println(">>> ", doc.Text())

	h, err := doc.Html()
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(">>> ", h)
}
```

This prints what you'd expect, without any script information. For this reason, I will decline the PR. Also, should you send other PRs in the future, note that tests are required unless the change is a trivial bug or doc fix.

Thanks and have fun with your project!
Martin

Dereking:
I got. Thanks.

PR #310, opened 2019-10-14
Updated colly project reference in README.md

theophanous:
asciimoo/colly has moved to gocolly/colly


mna:
Thanks, I'll try to see what's up with travis and Go 1.1.

PR #348, opened 2020-09-28
Send context to parseHtml

thiemok:
I stumbled upon the open bug in #178 and the open pull request to fix it in #235.
The original pull request seems to have been abandoned by its original creator, however I would like to have the original bug fixed.

This pr is a fork of the original one by @davidjwilkins and I have tried to address all the issues from the initial code review.

This is my first contribution to a go library so please let me know if I violated any conventions.
Let me know if you need any changes, or further documentation.

Original pr message for context:
```
This creates a new function parseHtmlWithContext, which behaves just like parseHtml but takes an *html.Node as context.

It also create a function eachNodeHtml which will iterate over each node in a selection, check to see if we have parsed html for this node type already, and if not, will call parseHtmlWithContext with either the current node, or it's parent, depending on whether you passed parent as true or false.

It also changes AfterHtml, AppendHtml, BeforeHtml, PrependHtml, ReplaceWithHtml, SetHtml, WrapHtml, WrapAllHtml, and WrapInnerHtml to use one or both of these functions.

New tests were added for all except the WrapHtml* functions - changing these functions may have been unncessary - I couldn't think of a way to test this - Go automatically wraps tr elements in tbody, so I can't wrap those, and I can't think of any other situation that this would occur in.
```

### DIFF
--- a/manipulation.go
+++ b/manipulation.go
@@ -39,8 +39,14 @@ func (s *Selection) AfterSelection(sel *Selection) *Selection {
 // AfterHtml parses the html and inserts it after the set of matched elements.
 //
 // This follows the same rules as Selection.Append.
-func (s *Selection) AfterHtml(html string) *Selection {
-	return s.AfterNodes(parseHtml(html)...)
+func (s *Selection) AfterHtml(htmlStr string) *Selection {
+	return s.eachNodeHtml(htmlStr, true, func(node *html.Node, nodes []*html.Node) {
+		for _, n := range nodes {
+			if node.Parent != nil {
+				node.Parent.InsertBefore(n, node.NextSibling)
+			}
+		}
+	})
 }
 
 // AfterNodes inserts the nodes after each element in the set of matched elements.
@@ -85,8 +91,12 @@ func (s *Selection) AppendSelection(sel *Selection) *Selection {
 }
 
 // AppendHtml parses the html and appends it to the set of matched elements.
-func (s *Selection) AppendHtml(html string) *Selection {
-	return s.AppendNodes(parseHtml(html)...)
+func (s *Selection) AppendHtml(htmlStr string) *Selection {
+	return s.eachNodeHtml(htmlStr, false, func(node *html.Node, nodes []*html.Node) {
+		for _, n := range nodes {
+			node.AppendChild(n)
+		}
+	})
 }
 
 // AppendNodes appends the specified nodes to each node in the set of matched elements.
@@ -123,8 +133,14 @@ func (s *Selection) BeforeSelection(sel *Selection) *Selection {
 // BeforeHtml parses the html and inserts it before the set of matched elements.
 //
 // This follows the same rules as Selection.Append.
-func (s *Selection) BeforeHtml(html string) *Selection {
-	return s.BeforeNodes(parseHtml(html)...)
+func (s *Selection) BeforeHtml(htmlStr string) *Selection {
+	return s.eachNodeHtml(htmlStr, true, func(node *html.Node, nodes []*html.Node) {
+		for _, n := range nodes {
+			if node.Parent != nil {
+				node.Parent.InsertBefore(n, node)
+			}
+		}
+	})
 }
 
 // BeforeNodes inserts the nodes before each element in the set of matched elements.
@@ -184,8 +200,12 @@ func (s *Selection) PrependSelection(sel *Selection) *Selection {
 }
 
 // PrependHtml parses the html and prepends it to the set of matched elements.
-func (s *Selection) PrependHtml(html string) *Selection {
-	return s.PrependNodes(parseHtml(html)...)
+func (s *Selection) PrependHtml(htmlStr string) *Selection {
+	return s.eachNodeHtml(htmlStr, false, func(node *html.Node, nodes []*html.Node) {
+		for _, n := range nodes {
+			node.InsertBefore(n, node.FirstChild)
+		}
+	})
 }
 
 // PrependNodes prepends the specified nodes to each node in the set of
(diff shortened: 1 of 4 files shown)


mna:
Hey, thanks for doing this!

Just wanted to let you know I have taken a quick look, it seems pretty good but I don't have time to do a thorough review to land this just yet - I'll do my best to get this done ASAP, hopefully before next week.

Thanks,
Martin

thiemok:
No worries, thats just fine with me. Thanks for the update!

manipulation.go
```diff
@@ -184,8 +200,12 @@ func (s *Selection) PrependSelection(sel *Selection) *Selection {
 }
 
 // PrependHtml parses the html and prepends it to the set of matched elements.
-func (s *Selection) PrependHtml(html string) *Selection {
-	return s.PrependNodes(parseHtml(html)...)
+func (s *Selection) PrependHtml(htmlStr string) *Selection {
+	return s.eachNodeHtml(htmlStr, false, func(node *html.Node, nodes []*html.Node) {
+		for _, n := range nodes {
+			node.InsertBefore(n, node.FirstChild)
```
mna, line 206: The order of the nodes to prepend would need to be reversed, otherwise this would insert the last node in nodes as first. I'll add a test to confirm what I think would happen.

manipulation.go
```diff
@@ -277,8 +304,17 @@ func (s *Selection) ReplaceWithNodes(ns ...*html.Node) *Selection {
...
 // specified html string.
-func (s *Selection) SetHtml(html string) *Selection {
-	return setHtmlNodes(s, parseHtml(html)...)
+func (s *Selection) SetHtml(htmlStr string) *Selection {
+	for _, context := range s.Nodes {
+		for c := context.FirstChild; c != nil; c = context.FirstChild {
+			context.RemoveChild(c)
+		}
+	}
+	return s.eachNodeHtml(htmlStr, true, func(node *html.Node, nodes []*html.Node) {
```
mna, line 313: Here I believe the isParent flag should be false, the node itself is the context for the html, not its parent. I'll also add a test to confirm this.

manipulation.go
```diff
@@ -334,8 +370,23 @@ func (s *Selection) WrapSelection(sel *Selection) *Selection {
...
+func (s *Selection) WrapHtml(htmlStr string) *Selection {
+	nodesMap := make(map[html.NodeType][]*html.Node)
+	var parent *html.Node
+	for _, context := range s.Nodes {
+		if context.Parent != nil {
+			parent = context.Parent
+		} else {
+			parent = &html.Node{Type: html.ElementNode}
+		}
+		nodes, found := nodesMap[parent.Type]
```
mna, line 382: The map (cache) should use the node name logic as in `eachNodeHtml` as key, not the Type.

manipulation.go
```diff
@@ -457,8 +518,17 @@ func (s *Selection) WrapInnerSelection(sel *Selection) *Selection {
 // cloned before being inserted into the document.
 //
 // It returns the original set of elements.
-func (s *Selection) WrapInnerHtml(html string) *Selection {
-	return s.wrapInnerNodes(parseHtml(html)...)
+func (s *Selection) WrapInnerHtml(htmlStr string) *Selection {
+	nodesMap := make(map[html.NodeType][]*html.Node)
+	for _, context := range s.Nodes {
+		nodes, found := nodesMap[context.Type]
```
mna, line 524: Same here, the key should be the node name.

manipulation.go
```diff
@@ -261,8 +281,15 @@ func (s *Selection) ReplaceWithSelection(sel *Selection) *Selection {
 // It returns the removed elements.
 //
 // This follows the same rules as Selection.Append.
-func (s *Selection) ReplaceWithHtml(html string) *Selection {
-	return s.ReplaceWithNodes(parseHtml(html)...)
+func (s *Selection) ReplaceWithHtml(htmlStr string) *Selection {
+	s.eachNodeHtml(htmlStr, true, func(node *html.Node, nodes []*html.Node) {
+		for _, n := range nodes {
+			if node.Parent != nil {
+				node.Parent.InsertBefore(n, node.NextSibling)
```
mna, line 288: Same here, I believe the order needs to be reversed to keep the original order of the html string intact.

manipulation.go
thiemok: I agree, must have totally skipped over those. I'll add the same logic as in `eachNodeHtml`

mna:
I can't push to the PR as it is from your fork and branch, so I pushed the changes I made to some tests and the helper loadString in the `thiemok-context` branch (https://github.com/PuerkitoBio/goquery/tree/thiemok-context). You can update your PR from it, or I believe you can give me push permissions to your fork following those instructions, as you wish: https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

thiemok:
I have done both and will clean up the rest later today.

mna:
Awesome, thanks!

thiemok:
Alright, I think I fixed everything. Please take a look at this fix here: https://github.com/PuerkitoBio/goquery/pull/348/commits/94c2530dd898fac51c676ea4a9f915ccaa00c2c8#diff-c3c4f0a035090792010e97ede68f785cR437
I think that was a typo in your test, but I want to make sure its actually asserting what you intended now

mna:
I added some test cases as the `AfterHtml` also had the ordering issue (I fixed it in my commit). The rest looks good to me, thanks a lot for your work! Note that I inadvertently pushed my commit to a new branch on your fork at first, I didn't realize my branch name was different than the PR's, you can ignore this.

I'll get this merged and cut a new release.

PR #356, opened 2020-11-06
add FindParent() and FindPrevious() with tests

fabstu:
I needed `FindParent(selector string)` and `FindPrevious(selector string)` functions for a small crawler project. I tried PrevUntil().Prev() to get the same result but that did not work, so I implemented them myself. I am uncertain how the functions should behave if Selection references more than 1 node.

Please let me know if I can improve anything.


mna:
Hello,

Thanks for caring about the package! Do you have a minimal, runnable example that demonstrates what you couldn't reproduce without those new functions? There's already the `Parents...` and `Prev...` family of functions that do very similar things, and it could be a documentation issue where it isn't clear what those functions do or how they work. I'd rather not add any API surface that doesn't match the jquery API.

Martin

fabstu:
I looked into it and the tests I added do not show the issue. Here is a more comprehensive example.

The following example steps into the link-element a, and then tries to find and print the first parent which matches the selector `div[class='il_ContainerListItem']`. See the full example below. I tried four approaches, and two worked for this case. 

```go
propertiesParent = link.ParentsUntil("div[class='il_ContainerListItem']").Parent().Last()
propertiesParent = FindParent(link, "div[class='il_ContainerListItem']")
```

The first is a reproduction using the current function, but it is non-intuitive that this is the same as "find first parent which matches and return it". This is why I think the new API, at least for `FindParent`, is justified. I did not look into FindPrevious but I think its the same there.


Full example:

```go
func TestFindParent4(t *testing.T) {
	testcase := `
<div class="il_ContainerListItem">
	<div class="il_ContainerListItem">
		<div class="il_ContainerItemTitle form-inline">
			<h4 class="il_ContainerItemTitle">
				<a href="abc" class="il_ContainerItemTitle">Discussion forum</a>
			</h4>
		</div>
	</div>
</div>
`

	doc, err := NewDocument(testcase)
	if err != nil {
		t.Fatal(err)
	}

	step1 := doc.Find("a.il_ContainerItemTitle")
	fmt.Println("step1",
		len(step1.Nodes),
		step1.Nodes[0].Data,
		step1.Nodes[0].Attr,
	)

	for i, node := range step1.Nodes {
		fmt.Println(i, ": step1:",
			len(step1.Nodes),
			node.Data,
			node.Attr,
		)
	}

	doc.Find("a.il_ContainerItemTitle").Each(
		func(n int, link *goquery.Selection) {
			if n != 0 {
				return
			}

			fmt.Println(strings.Repeat(fmt.Sprint(n), 10))

			propertiesParent := link.ParentsUntil("div[class='il_ContainerListItem']").Parent()

			fmt.Println("ParentsUntil().Parent():",
				len(propertiesParent.Nodes),
				propertiesParent.Nodes[0].Data,
				propertiesParent.Nodes[0].Attr,
				propertiesParent.Nodes[1].Data,
				propertiesParent.Nodes[1].Attr,
				propertiesParent.Text(),
			)

			propertiesParent = link.ParentsFiltered("div[class='il_ContainerListItem']")

			fmt.Println("ParentsFiltered():",
				len(propertiesParent.Nodes),
				propertiesParent.Nodes[0].Data,
				propertiesParent.Nodes[0].Attr,
				propertiesParent.Nodes[1].Data,
				propertiesParent.Nodes[1].Attr,
				propertiesParent.Text(),
			)

			propertiesParent = link.ParentsUntil("div[class='il_ContainerListItem']").Parent().Last()

			fmt.Println("ParentsUntil().Parent.Last():",
				len(propertiesParent.Nodes),
				propertiesParent.Nodes[0].Data,
				propertiesParent.Nodes[0].Attr,
				propertiesParent.Text(),
			)

			propertiesParent = findParent(link, "div[class='il_ContainerListItem']")

			fmt.Println("FindParent:",
				len(propertiesParent.Nodes),
				propertiesParent.Nodes[0].Data,
				propertiesParent.Nodes[0].Attr,
				propertiesParent.Text(),
			)
		},
	)
}
```

The shortened output:

```
ParentsUntil().Parent(): 2 div [{ class il_ContainerItemTitle form-inline}] div [{ class il_ContainerListItem}] 
                                Discussion forum
                                Discussion forum
ParentsFiltered(): 2 div [{ class il_ContainerListItem}] div [{ class il_ContainerListItem}] 
                                Discussion forum
                                Discussion forum
ParentsUntil().Parent.Last(): 1 div [{ class il_ContainerListItem}] 
                                Discussion forum
FindParent: 1 div [{ class il_ContainerListItem}] 
                                Discussion forum
```

mna:
Hello,

In that situation, what you want is `Closest`, it returns the first parent that matches the selector. Basically, the `Parent` (singular) family of functions get the immediate parent of the nodes in the selection (and matches the `parent` jquery function), `Parents` (plural) family of functions get all ancestors of the nodes in the selection (and matches the `parents` jquery function) and `Closest` get the first element of each nodes in the selection that matches the selector, starting from the node itself and traversing up (and matches the `closest` jquery function).

So this does what you want from your example:

```
const data = `
<div class="il_ContainerListItem">
	<div class="il_ContainerListItem">
		<div class="il_ContainerItemTitle form-inline">
			<h4 class="il_ContainerItemTitle">
				<a href="abc" class="il_ContainerItemTitle">Discussion forum</a>
			</h4>
		</div>
	</div>
</div>
`

func main() {
	doc, err := goquery.NewDocumentFromReader(strings.NewReader(data))
	if err != nil {
		log.Fatal(err)
	}
	parent := doc.Find("a").Closest("div.il_ContainerListItem")
	h, _ := goquery.OuterHtml(parent)
	fmt.Println(h)
}
```
(note that when searching by class, you can use the `.class` syntax). As mentioned, `Closest` starts with the selected node itself, so if you're worried that the node itself could match (impossible in that case as it's an `a` and we search for closest `div`), you should call `Parent()` before `Closest()`.

Hope this helps,
Martin

fabstu:
Hello.

Ahh I didn't find it, thank you! Is there the same for FindPrevious, which walks up with .Prev() and returns the first matching selector?

Anyway its good that I made this PR, now this is documented for people who search for FindParent. 👍

Cheers,
Fabian

mna:
There's not the equivalent of `Closest` for previous/next, but `PrevAllFiltered` is similar to `ParentsFiltered` in that it will add all previous siblings that match the selector. If you know only one of them match the selector, that's all you need, otherwise you may need some post-processing (e.g. call `First` afterwards, but that only works if operating on a base selection of one).

It would be nice to have some more general usage docs, in addition to the per-functions, as there are many functions to work through and the differences (or what they do) are not always obvious.

fabstu:
It is not obvious at all to call PrevAllFiltered().First, sadly. I agree that general docs can be improved. 👍 

I guess a mapping would be the most simple for people who know what they need from other languages. For example:
 
`jquery: prev(selector) -> PrevAllFiltered(selector).First() // if staring selection is one node, else have to postprocess.`

mna:
> It is not obvious at all to call PrevAllFiltered(sel).First()

I think it is well documented what `PrevAllFiltered` does, and from there you can infer that you may need to do more processing if you want to isolate a single previous sibling when many can match the selector, but what I mean by more general docs is docs that do not focus on API/function description, but a higher-level "if you have that kind of problem, use that kind of solution", because even if the function is documented, there are many of those and it's easy to miss that there is one that does what you need.

> I guess a mapping would be the most simple for people who know what they need from other languages.

I don't know soup (beautiful soup in Python, I assume? I know the name but that's pretty much all), but for jquery, the mapping is documented and straightforward. You may have misunderstood what I meant, `prev` in jquery works like `Prev` in goquery, and `PrevAll` is jquery's `prevAll`. I.e. there is no function in jquery that directly does `PrevAllFiltered(...).First()` either. I think the jquery mapping is quite well explained in the README: https://github.com/puerkitobio/goquery#api

PR #364, opened 2020-12-07
AddingPowerSupport_CI/Testing

santosh653:
Adding power support & updating the go versions to >=1.13 as lower versions are not supported.,

Adding power support ppc64le with Continues Integration/testing so that code remains architecture independent.

This is part of the Ubuntu distribution for ppc64le. This helps us simplify testing later when distributions are re-building and re-releasing. For more info tag @gerrith3.

The build is successful on both arch: amd64/ppc64le, please find the Travis Link below.
https://travis-ci.com/github/santosh653/goquery

Please let me know if you need any further details,

Thank You!!


mna:
Hello @gerrith3 ,

I have no problem adding test support for other architectures, but not at the expense of dropping tests on older versions of Go. There's surely a way to run a different set of Go versions for each architecture?

Martin

gerrith3:
@mna hey you can definitely customize the .travis.yml to run different things on different architectures.  @santosh653 should be able to update the .travis.yml.  Having said that, a bunch of those go versions should be pretty completely out of date.  The only versions I would really trust to be solid on ppc64le are 13/14/15 and later - earlier may work.  But you should be able to run any of the old versions on Intel if you want to.  Personally, I don't see a lot of value in that but you know your project better than we do.  And trimming the versions a little is a good way to save yourself some CI minutes, although that doesn't really matter - if you run out, ask travis for more.

mna:
Yeah let's not arbitrarily drop support for any old version on amd64, at least as long as travis or other CI let's us do so. Feel free to use just the versions you need on ppc.

santosh653:
@mna I have excluded what is not supported on Power ie go<=1.12.x & I am running the code for all the go versions for amd64.,
I hope this change in the code answers & addresses your query., 

The Travis Build is also successful with the new set of changes.
Looping @gerrith3

PR #370, opened 2021-01-11
Issue when parseHtmlWithContext is being called.

adam-cv:
Noticed a problem when using this package with advancedlogic/goose, a specific page was throwing a panic runtime error.

I changed the vendored code to log the page and context and noticed that a non ElementNode was being passed:
`panic(fmt.Sprintf("goquery: failed to parse HTML: %v , page: %v, context: %v", err, h, context))`

```
goquery: failed to parse HTML: html: ParseFragment of non-element Node , page: <p></p>, context: &{0xc000486460 0xc000487490 0xc000487f10 0xc0004873b0 0xc000487f80 1 span  <p></p>    []}
goroutine 68 [running]:
net/http.(*conn).serve.func1(0xc000798e60)
	/usr/local/opt/go/libexec/src/net/http/server.go:1801 +0x147
panic(0x18fe3c0, 0xc00489da50)
	/usr/local/opt/go/libexec/src/runtime/panic.go:975 +0x47a
github.com/PuerkitoBio/goquery.parseHtmlWithContext(0xc004822dc0, 0x9e, 0xc000487420, 0x5, 0x2271140, 0x0)
	.../vendor/github.com/PuerkitoBio/goquery/manipulation.go:585 +0x1df
github.com/PuerkitoBio/goquery.(*Selection).eachNodeHtml(0xc001478c60, 0xc004822dc0, 0x9e, 0x1, 0x1ab5088, 0x9e)
	.../vendor/github.com/PuerkitoBio/goquery/manipulation.go:675 +0x1d0
github.com/PuerkitoBio/goquery.(*Selection).BeforeHtml(0xc001478c60, 0xc004822dc0, 0x9e, 0xc004822d20)
	.../vendor/github.com/PuerkitoBio/goquery/manipulation.go:139 +0x50
github.com/advancedlogic/GoOse.(*Cleaner).convertDivsToParagraphs.func1(0x1, 0xc001478900)
	.../vendor/github.com/advancedlogic/GoOse/cleaner.go:553 +0x298
```

Not sure why goose passes a non html.ElementNode to this function, but the panic is not great.

Added a test case for this.

I'm not sure this is the correct fix as it's not clear why parseHtmlWithContext is being called for a non element node.

But creating the PR incase someone else runs into this problem.


mna:
Hello,

Thanks for raising this, there is a possibility of a panic when calling Append/PrependHtml on a Selection that contains non-Element nodes (e.g. a Selection created via `.Contents()`). I will push a fix for this - to ignore non-Element nodes, which is how jquery behaves in this case - and cut a new release.

Martin

PR #379, opened 2021-04-20
Implement `Render` top-level function

### LINKED ISSUE #377: Is it possible to save html that was manipulated using `goquery`?
If I understand correctly there are few methods that allow changing of HTML structure such as `Add*` and `Remove*`. I would like to change loaded document and then save it back to say, a file, or  using a `io.Writer` interface. Is it possible?

inliquid:
Fixes #377

### DIFF
--- a/utilities.go
+++ b/utilities.go
@@ -80,9 +80,9 @@ func OuterHtml(s *Selection) (string, error) {
 
 // Render renders first item in the selection (like in OuterHtml - that is,
 // the HTML including the first element's tag and attributes) to the given writer
-// by calling `html.Render`. In case of empty `s`, empty byte slice will be written.
+// by calling html.Render. In case of empty s, empty byte slice will be written.
 //
-// In order to render entire document, pass a `doc.Selection` as `s`.
+// In order to render entire document, pass a doc.Selection as s.
 func Render(w io.Writer, s *Selection) error {
 	if s.Length() == 0 {
 		_, err := w.Write([]byte{})
--- a/utilities.go
+++ b/utilities.go
@@ -2,6 +2,7 @@ package goquery
 
 import (
 	"bytes"
+	"io"
 
 	"golang.org/x/net/html"
 )
@@ -75,6 +76,20 @@ func OuterHtml(s *Selection) (string, error) {
 		return "", err
 	}
 	return buf.String(), nil
+}
+
+// Render renders first item in the selection (like in OuterHtml - that is,
+// the HTML including the first element's tag and attributes) to the given writer
+// by calling `html.Render`. In case of empty `s`, empty byte slice will be written.
+//
+// In order to render entire document, pass a `doc.Selection` as `s`.
+func Render(w io.Writer, s *Selection) error {
+	if s.Length() == 0 {
+		_, err := w.Write([]byte{})
+		return err
+	}
+	n := s.Get(0)
+	return html.Render(w, n)
 }
 
 // Loop through all container nodes to search for the target node.
--- a/utilities_test.go
+++ b/utilities_test.go
@@ -10,10 +10,8 @@ import (
 	"golang.org/x/net/html"
 )
 
-var allNodes = `<!doctype html>
-<html>
-	<head>
-		<meta a="b">
+const allNodes = `<!DOCTYPE html><html><head>
+		<meta a="b"/>
 	</head>
 	<body>
 		<p><!-- this is a comment -->
@@ -22,8 +20,7 @@ var allNodes = `<!doctype html>
 		<div></div>
 		<h1 class="header"></h1>
 		<h2 class="header"></h2>
-	</body>
-</html>`
+</body></html>`
 
 func TestNodeName(t *testing.T) {
 	doc, err := NewDocumentFromReader(strings.NewReader(allNodes))
@@ -143,6 +140,7 @@ func TestRender(t *testing.T) {
 	nText := nComment.NextSibling
 	sHeaders := doc.Find(".header")
 	sEmpty := doc.Find(".empty")
+	sFull := doc.Selection
 
 	cases := []struct {
 		node *html.Node
@@ -159,7 +157,8 @@ func TestRender(t *testing.T) {
 		This is some text.
 		`},
 		{nil, sHeaders, `<h1 class="header"></h1>`},
-		{nil, sEmpty, ""}, // empty selection
+		{nil, sEmpty, ""},      // empty selection
+		{nil, sFull, allNodes}, // full document selection
 	}
 	for i, c := range cases {
 		if c.sel == nil {
--- a/utilities_test.go
+++ b/utilities_test.go
@@ -1,6 +1,7 @@
 package goquery
 
 import (
+	"bytes"
 	"reflect"
 	"sort"
 	"strings"
@@ -120,6 +121,58 @@ func TestOuterHtml(t *testing.T) {
 		if err != nil {
 			t.Fatal(err)
 		}
+
+		if got != c.want {
+			t.Errorf("%d: want %q, got %q", i, c.want, got)
+		}
+	}
+}
+
+func TestRender(t *testing.T) {
+	doc, err := NewDocumentFromReader(strings.NewReader(allNodes))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	n0 := doc.Nodes[0]
+	nDT := n0.FirstChild
+	sMeta := doc.Find("meta")
+	sP := doc.Find("p")
+	nP := sP.Get(0)
+	nComment := nP.FirstChild
+	nText := nComment.NextSibling
+	sHeaders := doc.Find(".header")
+	sEmpty := doc.Find(".empty")
+
+	cases := []struct {
+		node *html.Node
+		sel  *Selection
+		want string
+	}{
+		{nDT, nil, "<!DOCTYPE html>"}, // render makes DOCTYPE all caps
+		{nil, sMeta, `<meta a="b"/>`}, // and auto-closes the meta
+		{nil, sP, `<p><!-- this is a comment -->
+		This is some text.
+		</p>`},
+		{nComment, nil, "<!-- this is a comment -->"},
+		{nText, nil, `
+		This is some text.
+		`},
+		{nil, sHeaders, `<h1 class="header"></h1>`},
+		{nil, sEmpty, ""}, // empty selection
+	}
+	for i, c := range cases {
+		if c.sel == nil {
+			c.sel = newSingleSelection(c.node, doc)
+		}
+
+		buf := &bytes.Buffer{}
+
+		err := Render(buf, c.sel)
+		if err != nil {
+			t.Fatal(err)
+		}
+		got := buf.String()
 
 		if got != c.want {
 			t.Errorf("%d: want %q, got %q", i, c.want, got)


utilities.go
```diff
@@ -77,6 +78,20 @@ func OuterHtml(s *Selection) (string, error) {
 	return buf.String(), nil
 }
 
+// Render renders first item in the selection (like in OuterHtml - that is,
+// the HTML including the first element's tag and attributes) to the given writer
+// by calling `html.Render`. In case of empty `s`, empty byte slice will be written.
+//
+// In order to render entire document, pass a `doc.Selection` as `s`.
```
mna, line 85: Small nit: Go's comments do not use markdown syntax, the community convention is to just use standard text (i.e. like the standard library's documentation). Let's remove the backticks.

utilities.go
```diff
@@ -77,6 +78,20 @@ func OuterHtml(s *Selection) (string, error) {
...
+func Render(w io.Writer, s *Selection) error {
+	if s.Length() == 0 {
+		_, err := w.Write([]byte{})
```
mna, line 88: Any reason why you make an empty `Write` call when the selection is empty? I'd simply do `return nil` here.

utilities_test.go
```diff
@@ -126,3 +127,55 @@ func TestOuterHtml(t *testing.T) {
...
+		{nil, sP, `<p><!-- this is a comment -->
+		This is some text.
+		</p>`},
+		{nComment, nil, "<!-- this is a comment -->"},
+		{nText, nil, `
+		This is some text.
+		`},
+		{nil, sHeaders, `<h1 class="header"></h1>`},
+		{nil, sEmpty, ""}, // empty selection
+	}
```
mna, line 163: Would be nice to do a full document rendering, just to make sure (i.e. sel = doc.Selection).

mna (changes requested):
Hey, thanks for this! It looks good, I only have a few minor comments, let me know if you don't have time to address them, I can do them before merging your changes to main.

utilities.go
inliquid: Yep. I wasn't sure about this, fixed.

utilities.go
inliquid: Yes, what I thought of was that as `Render` returns an error, I would expect to see any error coming from `io.Writer` even when there is empty value to be written. In fact, caller may not know what is going to be written, will the value become empty or not. However, if socket has been closed, `EOF` reached or whatsoever, I want `Render` to fail. Otherwise it would be surprising if f.e. Render returned `nil` as there was no error, and next attempt to write will fail because something is actually wrong with the writer.

utilities_test.go
inliquid: Added. However I had to slightly change `allNodes`, as `Render` seems to loose some of formatting and capitalize "doctype". I also made it const as there seems to be no reason to keep it variable.

utilities.go
mna: I don't know, I feel like this is trying to be overly clever, to me it would be more surprising to have render fail when there was nothing to actually write (empty selection), and it's something the caller can verify prior to calling Render if it is important. It's not the same thing as making a direct call to Write to the Writer, it's a higher-level function and it may not have anything to send, so it makes sense to me to return nil in that case without attempting a Write. If there _is_ a selection, then abolutely, it should make a Write call - and it will. For example - it's not a perfect comparison, but I think it's a relevant one - even an explicit call to bufio.Writer.Flush() will not call Write if there is no data to write, it will just return nil.

I'd prefer to return nil here, and to mention in the function's documentation that no Write attempt is made if the selection is empty.

utilities_test.go
mna: :+1: cool, yeah that's fine, the html rendering "canonicalizes" the representation and whitespace is not significant between the html tag and head/body so it is lost. Good call for the `const`.

utilities.go
inliquid: >even an explicit call to bufio.Writer.Flush() will not call Write if there is no data to write

Yes, but not exactly the same. `Flush` just flushes the buffer, and it returns `nil` if there is nothing left, that's correct, however, beforehand it checks for writer error state:
https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/bufio/bufio.go#L601

And then, when if it calls `Write`, it returns any error, and even creates new when amount of written bytes is less than in buffer:
https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/bufio/bufio.go#L608

If `Render` will not return any error on attempt to write to broken `Writer`, than you could imagine making that check explicitly in addition to calling `Render`, if that for some reason is important for the program.

For instance:
```go
if err := Render(w, s); err != nil {
    <...>
} else if s.Length() == 0 && err == nil {
    // Check that writer is still fine
    _, err := w.Write([]byte{}) {
        <...>
    }
}
```

What if instead of ignoring current state of writer we wrap that error to reflect exactly what the problem is?
```go
_, err := w.Write([]byte{})
return fmt.Errorf("writer not ready: %w", err)
```

mna: I disagree, it shouldn't touch the `Writer` if there is nothing to write. Here is perhaps a more convincing example, with the behaviour of `io.Copy`:

```
type PanicWriter int

func (p PanicWriter) Write(b []byte) (int, error) {
	panic("written to")
}

func main() {
	var pw PanicWriter
	fmt.Println(io.Copy(pw, strings.NewReader("")))
}
```

And in the case of goquery, it's even more in favour of not touching the writer as you can actually easily check if it needs to write anything beforehand, which is not as easy with a Reader.

inliquid: > I disagree, it shouldn't touch the Writer if there is nothing to write.

It __must__ touch `Writer` because that's exactly what it does - __writes__ rendered content. 

I suppose that check for selection emptiness is because if it's empty and there is no check, it will raise a panic in this:
```go
...
n := s.Get(0)
...
```
because `Get` just indexes what's in selection:
```go
...
return s.Nodes[index]
``` 

But if there was no panic, would you agree to reimplement it like this?
```go
func Render(w io.Writer, s *Selection) error {
	n := s.Get(0)
	return html.Render(w, n)
}
```
So maybe `Get` should instead check for nilness before accessing the slice to avoid painic?

Another approach to handle whole situation, if you don't want to touch the writer would be like this:
```go
func Render(w io.Writer, s *Selection) error {
	if s.Length() == 0 {
		return errors.New("empty selection")
	}
	n := s.Get(0)
	return html.Render(w, n)
}
```

mna: Please don't ignore what I just mentioned, otherwise we're just losing time arguing past each other's point - `io.Copy` does exactly what I propose, it doesn't go out of its way to make a call to Write to check for Writer's state if there is no byte to write. `goquery.Render` is a higher-level function whose purpose is not to check if your `io.Writer` is still alive, that's what keep-alives, heartbeats and similar are for if this is needed (which is not the general case), and otherwise any actual bytes to write will tell you (and if there are no bytes to write, it won't tell you because in most cases it doesn't matter - there was nothing to write anyway - you'll know on the next write or maybe on close).

Also for consistency with the rest of goquery's api, it's not an error to call a function with an empty selection, e.g. `OuterHtml` returns `"", nil`, and `NodeName` returns empty string. 

For those reasons I don't see any compelling reason not to have this be:

```
func Render(w io.Writer, s *Selection) error {
	if s.Length() == 0 {
		return nil
	}
	n := s.Get(0)
	return html.Render(w, n)
}
```

And mention that no write is done if there is no Selection just to be explicit about it.

inliquid: Your example of `io.Copy` is incorrect, because `io.Copy` does not _write_, it _reads_ and _writes_ (it's actually a little more complicated under the hood, with buffering and asserting `io.ReaderFrom` and `io.WriterTo`). As I already said in my previous comments, caller of `Render` does not necessarily know what he is going to write, if selection is empty or not. But he will expect same behavior of `Render` for same (presumably broken) writer, no matter if it's empty or not. If you'd offered to implement `CopyHTML` procedure analogue to `io.Copy`, I'd agree with what you say.

inliquid: > Also for consistency with the rest of goquery's api, it's not an error to call a function with an empty selection, e.g. OuterHtml returns "", nil, and NodeName returns empty string.

`Render` and `OuterHTML` are different functions which can error differently?

mna: The analogy is pretty good, you're basically playing with words, if you prefer to read it that way so that words match, `io.Copy` "renders" the reader to the writer, and does not write anything if there's nothing to write. That's exactly what is going on with `goquery.Render` (in fact, it's higher-level than io.Copy and makes even more sense to *not render anything* if there is nothing to render). Again, its purpose is not to check for the health of your writer. I've argued well past the reasonable amount of time I was expecting to spend on this issue, so honestly my decision is made by now.

Let me know if you're ok with me adjusting the code the way I proposed before merging, as I don't want to force you to contribute your code if you're angry with the direction.

inliquid: `io.Copy` which consumes both `Reader` and `Writer` does not _write_, it _copies_ data from reader to writer which is approximately what's meant by words "it _reads_ and _writes_". `Copy` "knows" when there is nothing to write, which is not the case for direct `Write` call or `Render` call. This is not playing with words, but plain semantics of operations.

Consider simple case. There is a `Writer` which returns `io.EOF` on any attempt to write anything. And there are selections `s`, `s1`, and `s2`:
```go
// s is not empty
// s1 is empty
// s2 not empty

err := goquery.Render(w, s)
if err != nil { // io.EOF, maybe you need new writer?
     handleRenderErr(err)
}

err = goquery.Render(w, s1)
if err != nil { // nil, no errors?
     handleRenderErr(err) // never reached
}

err = goquery.Render(w, s2)
if err != nil { // io.EOF again??
     handleRenderErr(err)
}
```

So this difference in behavior implies that developer have to always assert and workaround empty selection case. Which in fact means that `Render` should not handle empty selection at all, and just return an error, at least this looks valid. Or do what was proposed initially, translate `Writer` error by writing an empty slice.

>I've argued well past the reasonable amount of time I was expecting to spend on this issue, so honestly my decision is made by now.

Sorry but this is not how engineering works, like it or not. If issue requires more time, than you've expected, you either spend it, or deliver broken things.

mna: > Copy "knows" when there is nothing to write, which is not the case for direct Write call or Render call. 

I've never said it was wrong for direct `Write` call, so let's leave that out, but for `Render`, it trivially "knows" when there is nothing to write - that's when the Selection is empty.

> Consider simple case.

Consider the same simple case with `io.Copy` and various Readers, some empty. Same thing, you won't know if, as you put it, you "need a new writer".

Listen, I get it, you disagree with my decision and made that pretty clear. I disagree with your proposal - to reiterate for the nth time it is not the job of that function to validate your Writer's health. Couple more things: a) No need to make it absolute and insulting, we can disagree without trying to make the other look like they don't know a thing, and b) be nice to open-source maintainers, nobody owes you their time.

Last time I ask, are you ok with me adjusting the code before merging or you don't want to contribute this PR?

inliquid: What you observed with `io.Copy` is not its behavior, in fact, as I told, it asserts `WriterTo` on `Reader` or `ReaderFrom` on `Writer` and totally relies on them. In the above case you don't see `Write` happening, because that is how `strings.Reader` is implemented.

Here is your panicking `io.Copy`:
```go
package main

import (
	"fmt"
	"io"
)

type PanicWriter struct{}

func (*PanicWriter) Write(b []byte) (int, error) {
	panic("written to")
}

type EmptyReader struct{} // implements io.WriterTo

func (*EmptyReader) Read(b []byte) (int, error) {
	return 0, io.EOF
}

func (*EmptyReader) WriteTo(w io.Writer) (int64, error) {
	n, err := w.Write([]byte{})
	return int64(n), err
}

func main() {
	pw := &PanicWriter{}
	er := &EmptyReader{}

	fmt.Println(io.Copy(pw, er))
}
```
Output:
```
panic: written to

goroutine 1 [running]:
main.(*PanicWriter).Write(0x44ee00, 0x44ee00, 0x0, 0x0, 0x359d80, 0x1, 0x1ca3b99d958)
...
```

inliquid: >a) No need to make it absolute and insulting, we can disagree without trying to make the other look like they don't know a thing, and 

I never told you anything insulting in any of my comments, like "Last time I ask".

>b) be nice to open-source maintainers, nobody owes you their time

Why do you think that your and only your time matters? Did I owe you my time?

>Last time I ask,...

Ok, that's too much. This is not how a "maintainer" should face contributor, or anyone. I never gave an opportunity to talk to me with this tone. 

I'm done here, and **I do not provide permission to use my code in any part of your project**.

mna: Cool. For future/other contributors interested in adding this, note that it's hardly surprising that making an explicit call to `PanicWriter.Write` panics...

This example is more interesting, and as a bonus adds other examples of higher-level "Write"-like functions that don't make a call to the base writer when there's nothing to write:

```
type PanicWriter int

func (p PanicWriter) Write(b []byte) (int, error) {
	panic("written to")
}

type NotWriterToReader int

func (n NotWriterToReader) Read(b []byte) (int, error) {
	return 0, io.EOF
}

func main() {
	var pw PanicWriter
	var nr NotWriterToReader
	fmt.Println(io.Copy(pw, nr))

	cw := csv.NewWriter(pw)
	fmt.Println(cw.WriteAll(nil))

	bw := bufio.NewWriter(pw)
	fmt.Println(bw.Write([]byte{}))
	fmt.Println(bw.Flush())
}
```

Also, and perhaps more importantly, in real-life cases, making a dummy empty call to the writer will almost never give you anything useful, as writers that may be candidates to return "health information" via a Write error (like a File or a socket) would typically be wrapped in a buffered writer to avoid unnecessary syscalls, so making an empty write to a buffer does not trigger any write to the base write, as shown in my last example.

Martin

inliquid: > Also, and perhaps more importantly, in real-life cases, making a dummy empty call to the writer will almost never give you anything useful, as writers that may be candidates to return "health information" via a Write error (like a File or a socket) would typically be wrapped in a buffered writer to avoid unnecessary syscalls, so making an empty write to a buffer does not trigger any write to the base writer, as shown in my last example.

Well that is just false. Please come back when you experience more real-world cases, such as timed-out `http.Writer` or just this simple thing:
```go
func main() {
	f, err := os.Create("test.txt")
	if err != nil {
		panic(err)
	}

	if _, err := f.Write([]byte("text")); err != nil {
		panic(err)
	}

	f.Close()

	fmt.Println(f.Write([]byte{}))
}
```

PR #381, opened 2021-06-04
Fixed ExampleScrape README Sample

j0nimost:
Changed the README `ExampleScrape() ` example slightly which now works with [metalsucks.net](http://metalsucks.net). Below is the output.

image


mna:
Awesome, thanks!

PR #386, opened 2021-07-21
Add GetSingleSelectionList

SirRFI:
Sometimes I encounter scenarios where `Each` method is insufficient, for example when I need to return an error during the iteration. One of first things I tried was looping through `selection.Nodes`, but working with that brings it's own issues as far I remember. So, I ended up doing this:
```go
func GetSingleSelectionList(selection *goquery.Selection) []*goquery.Selection {
	selectionList := make([]*goquery.Selection, selection.Length())
	selection.Each(func (i int, sel *goquery.Selection) {
		selectionList[i] = sel
	})
	return selectionList
}
```
This feels like doing something that should be built in though.

This is where this PR comes in. I am not sure how to name this method though - for now it's `GetSingleSelectionList`, unless it conflicts with existing naming. Maybe something with `split` or `spread` wording would fit better, or go other direction and make it more accessible.

### DIFF
--- a/iteration.go
+++ b/iteration.go
@@ -37,3 +37,13 @@ func (s *Selection) Map(f func(int, *Selection) string) (result []string) {
 
 	return result
 }
+
+// GetSingleSelectionList iterates over a Selection object and returns all
+// matched elements as a slice of single-element Selection.
+func (s *Selection) GetSingleSelectionList() []*Selection {
+	list := make([]*Selection, len(s.Nodes))
+	for i, n := range s.Nodes {
+		list[i] = newSingleSelection(n, s.document)
+	}
+	return list
+}
--- a/iteration_test.go
+++ b/iteration_test.go
@@ -86,3 +86,29 @@ func TestForRange(t *testing.T) {
 		t.Errorf("expected initial selection to still have length %d, got %d", initLen, sel.Length())
 	}
 }
+
+func TestGetSingleSelectionList(t *testing.T) {
+	expectedValues := []string{"n1", "n2", "n3", "n4", "n5", "n6"}
+
+	sel := Doc2().Find("#main div")
+	result := sel.GetSingleSelectionList()
+
+	if len(result) != 6 {
+		t.Errorf("Expected GetSingleSelectionList result to have a length of 6, found %v.", len(result))
+	} else {
+		for i, s := range result {
+			if s.Length() != 1 {
+				t.Errorf("Expected result[%d] to contain only one selected element, got %d", i, s.Length())
+			}
+
+			value, exists := s.Attr("id")
+			if exists == false {
+				t.Fatalf("Expected result[%d]'s node to have \"id\" attribute", i)
+			}
+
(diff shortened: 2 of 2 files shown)


mna:
Hello,

I'm not sure I understand the purpose this new method would serve, what you are building is a slice of single-element selections which is more or less what the original selection is. What would you do then with that slice of single selections? If it's just to return an error from `Each` as you mention, this is easy to accomplish by assigning to an error variable declared outside the Each (or EachWithBreak) closure.

If you could provide a representative task that you struggled to accomplish with the current API, it would help illustrate how adding this would help or perhaps I could see if a different approach using the (quite large) existing API would be better.

Martin

SirRFI:
I am currently rewriting my scrapping program and managed to solve various problems differently than I initially did. This one is keep returning though. Unfortunately I don't remember all the scenarios I experienced through the way. Currently I have two where it's used:
1. Looping through a list and performing some actions, which can return an error
2. Looping through another list backwards

For the first one, your suggestion indeed solved the issue. I should have thought of that, especially that I already used non-error variables like that. For my defense, I am relatively new to Go.

For the second one I still need a slice of selections. The reason I go backwards on it is because I obtain the HTML, do some work on it and set it back to the selection. Doing it in normal order may invalidate further selections if they were found in that HTML, while doing it backwards seems to start from youngest/most nested/indented ones and then move to their ancestors. I am uncertain yet if this is reliable, it's a fresh idea. `div` can come into mind as example, but I find `ul` better one, as `li` elements are between layers - to make nested `ul`, it has to be in `li` element. Thoughts?

mna:
Yeah if you change the html of elements in the selection, it can definitely invalidate elements from the selection if some were nested inside the modified element. That being said, there are a number of html manipulation methods you can use that can append/prepend/wrap etc html on the selection as a whole, if the manipulation you have to do is the same for all nodes in the selection.

Otherwise, you can still do it in whatever order (assuming you don't have to re-select and re-process elements inside the newly assigned html?), because replaced elements will still be valid in the selection, it's just that they won't be part of the document anymore (e.g. if you print out the outer html of the root document, they won't appear, they are orphan nodes).

You cannot really rely on the order in a selection in general, because a selection can be created from many different ways (e.g. Add other selections, Intersection, Union, etc.) that can change the ordering. But for a single operation (e.g. doc.Find("div")), then yes, the document order is respected, so parent nodes will appear before nested ones in the selection.

If you really want to iterate in reverse order, you can use something like that (untested):

```
for i := sel.Length() - 1; i >= 0; i-- {
   singleSel := sel.Eq(i)
   // use selection...
}
```

Hope this helps,
Martin

SirRFI:
Your suggestion seems to do the trick indeed. I expected that if such feature wasn't added this long, there are probably ways to do it somehow else. Well, in case someone runs into similar problem / requirement, this discussion may be in use. Thank You your time, help and the library.

mna:
Cool, no problem, thanks for your interest in goquery and have fun with your project!

PR #392, opened 2021-09-22
use new Cascadia version

kinoute:
The new Cascadia version includes:

- Case-insensitive CSS selectors attributes with `i`(https://github.com/andybalholm/cascadia/pull/51)
- Support of a lot of pseudo-classes (https://github.com/andybalholm/cascadia/issues/50)


mna:
Thanks!

PR #393, opened 2021-09-23
feat: return selections

chyroc:
add `.Selections()` to `*Selection`, so i can use lambda functions to handle it

example:

```go
import "github.com/chyroc/go-lambda"

func main()
	sels := Doc().Find(".hero-unit .row-fluid").Selections()

	lambda.New(sels).ArrayAsync(func(idx int, obj interface{}) interface{} {
		// do http request with async, faster than .Each
		http.Get(obj.(*Selection).AttrOr("href", ""))
		return obj.(*Selection).Text()
	})
}
```

### DIFF
--- a/iteration.go
+++ b/iteration.go
@@ -1,5 +1,14 @@
 package goquery
 
+// Selections returns each element in the selection as Selection list
+func (s *Selection) Selections() []*Selection {
+	res := make([]*Selection, len(s.Nodes))
+	for i, n := range s.Nodes {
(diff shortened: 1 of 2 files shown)


mna:
Hello,

Thanks for your interest in goquery, but I don't see a compelling argument to add this - one can easily collect the selections via `Each` or use `Map` to run transformations directly, while there's also the `*Selection.Nodes` slice that contains the raw `*html.Node` values.

Thanks,
Martin

PR #404, opened 2022-01-04
Updated go.mod dependencies per Snyk alert

jauderho:
Signed-off-by: Jauder Ho <jauderho@users.noreply.github.com>

Snyk identified goquery as needing to have an updated golang.org/x/net/http2. See https://security.snyk.io/vuln/SNYK-GOLANG-GOLANGORGXNETHTTP2-2313688

This PR just bumps the x/net version.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/PuerkitoBio/goquery
 
 require (
 	github.com/andybalholm/cascadia v1.3.1
-	golang.org/x/net v0.0.0-20210916014120-12bc252f5db8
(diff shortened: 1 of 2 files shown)


mna:
Thanks!

mna:
Just a note for others stumbling on this, goquery does not use the `http2` package and as such was not impacted by the security alert. It only uses the `golang.org/x/net/html` package, but since both `http2` and `html` are in the same repo, it probably got flagged as requiring an upgrade. No harm done either way, not a bad thing to update the deps!

PR #417, opened 2022-06-16
add suppert slice text

zkep:
such as get breadcrumbs navigation

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -108,3 +108,31 @@ func ExampleSingle() {
 	// 123
 	// 1
 }
+
+func ExampleSeparatorText() {
+	html := `
+<html>
+  <body>
+    <div>1</div>
+    <div>2</div>
+    <div>3</div>
+  </body>
+</html>
+`
+	doc, err := goquery.NewDocumentFromReader(strings.NewReader(html))
+	if err != nil {
+		log.Fatal(err)
+	}
+	separator := ","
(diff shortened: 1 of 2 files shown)


mna:
Hello!

Thanks for your interest in goquery! As a general rule, you should always start by discussing new features in an issue instead of sending a Pull Request directly - it can avoid misunderstandings and unnecessary time spent working on a feature that doesn't get accepted.

That being said, I'm not totally against the idea, but you know how it goes, "yes" is forever (it has to be maintained in the package "forever", until a new major version is released with breaking changes/deprecation) while "no" can be temporary, and at the moment, I don't remember anyone asking for such a feature in goquery in past issues, so I'm not convinced it is worth adding to the package.

Also, goquery tries to stay close to the jquery API and this is not a jquery method, so if it was to be accepted, it would need to be moved from a method on the `Selection` type to a package-level function that takes a `Selection` as argument (e.g. a bit like other non-jquery-compatible functions, such as https://pkg.go.dev/github.com/PuerkitoBio/goquery#NodeName).

All this to say, for the moment I don't feel it's worth adding to goquery. You can do something similar if you don't have a deeply nested set of elements with this:

```
	sl := multiSel.Map(func(i int, s *goquery.Selection) string {
		return s.Text()
	})
	fmt.Println(sl)
```

But of course if you have nested elements, this would not quite work.

Martin

PR #427, opened 2022-09-03
Update installation instructions

noah-de:
Otherwise new comers who want to follow along would see something like this:
        'go get' is no longer supported outside a module.
	To build and install a command, use 'go install' with a version,
	like 'go install example.com/cmd@latest'
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Syntax-wise, it is as close as possible to jQuery, with the same function names 
 
 Please note that because of the net/html dependency, goquery requires Go1.1+ and is tested on Go1.7+.
 
-    $ go get github.com/PuerkitoBio/goquery
+    $ go install github.com/PuerkitoBio/goquery@latest
 
 (optional) To run unit tests:
 


mna:
`go install` with the version specifier is to install commands (executables). Goquery is a package, `go get` is the right way to install it, it's just that you now have to be in a Go module to install it. I'd welcome a PR that makes that clearer.

noah-de:
Thanks for the kind reply. 
I am still getting used to how things work in Go, (recognizing a project's purpose via the structure of the source code).
Now I see that Goquery is a package installed (as a source based dependency) and then imported into code that is being written. The documentation does a pretty good job of showing that.
I was aware of the official discussion, https://github.com/golang/go/issues/40276  but am still new enough to the language that misunderstood the meaning, until now.

Additionally, I wanted to run the tests as well (according to instructiuons in the readme) but 
 `cd $GOPATH/src/github.com/PuerkitoBio/goquery` doesn't exist, since the install happened was for the module that I am working in and not the whole computer (as it would with an installed command).

A little digging (`go env`) revealed that a valid test path could be found in $GOMODCACHE: 
`go test ~/go/pkg/mod/github.com/\!puerkito\!bio/goquery@v1.8.0/`

However, that doesn't appear to be something that we would direct folks to run to test this code.
Do you have any suggestions for the correct way to run tests in the code that was installed with `go get`?

Many thanks!

PR #469, opened 2024-02-28
Pre-allocate slice in generic Map function

jwilsson:
Hey!
Thanks for this great library, really helped me out with some scraping!

I noticed that the generic `Map` function just added the other week always `append`s, growing the slice as needed instead of pre-allocating it upfront.

Benchmark results from my M1:
```
Without pre-allocation:
BenchmarkMap-10                              	  238780	      5140 ns/op

With pre-allocation:
BenchmarkMap-10                              	  277968	      4317 ns/op
```

If you think it's not worth it, no hard feelings. Just thought it would be a fun, easy, optimization to make.

### DIFF
--- a/iteration.go
+++ b/iteration.go
@@ -37,8 +37,10 @@ func (s *Selection) Map(f func(int, *Selection) string) (result []string) {
 // Map is the generic version of Selection.Map, allowing any type to be
 // returned.
 func Map[E any](s *Selection, f func(int, *Selection) E) (result []E) {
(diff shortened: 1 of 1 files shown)


mna:
Hey Jonathan,

Thanks for this! That behaviour actually predates the new `Map` addition as this is also how I had implemented the `Selection.Map` method. That's definitely worth it, I'll get that merged soon.

Martin

PR #478, opened 2024-07-02
Bump GitHub workflows to their latest versions

deining:
This PR bumps GitHub workflows to their latest versions, thus avoiding deprecation warnings as seen [here](https://github.com/PuerkitoBio/goquery/actions/runs/9439105361).


mna:
Thank you!

PR #483, opened 2024-07-20
Support IterEach based on 1.23 Iterator

### LINKED ISSUE #482: Question: Is it possible support the iterator in version 1.23?
Golang 1.23 will support the `iter` package and define the types `iter.Seq` and `iter.Seq2` as iterators. My idea is that maybe `Each` can return an iterator. Then the user can use the `for loop` on it easily. 
```go
func (s *Selection) EachIter() iter.Seq2[int, *Selection]
```

At client side:
```go
for i, s := range doc.Find(".left-content article .post-title").Each() {
    title := s.Find("a").Text()
    fmt.Printf("Review %d: %s\n", i, title)
}
```

# Reference
https://tip.golang.org/doc/go1.23
https://pkg.go.dev/iter@master

amikai:
Resolve #482

# What's change
- Bump to golang 1.23
- Support `IterEach`

Note: Considering that Go 1.23 is not stable yet, I am not sure if merging that PR would be risky.

# Detail

Test cases:
`TestEachIter` is similar with `TestEach`
`TestEachIterWithBreak` is similar with`TestEachWithBreak`

Benchmarks:
`BenchmarkEachIter` is similar with `BenchmarkEach`
`BenchmarkEachIterWithBreak` is similar with `BenchmarkEachWithBreak`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.21.x, 1.22.x, 1.23rc2]
+        go-version: [1.21.x, 1.22.x, 1.23.0-rc.2]
         os: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.21.x, 1.22.x, 1.23.0-rc.1]
+        go-version: [1.21.x, 1.22.x, 1.23rc2]
         os: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.21.x, 1.22.x]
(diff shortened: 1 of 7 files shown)


mna (commented):
The PR looks great, thanks for your contribution! As you mention in the description, go1.23 is not stable yet so I'll just wait for it to be released before merging and cutting a new release. Don't hesitate to ping me when 1.23 is released if you see no progress on that front, just as a reminder to merge this!

amikai:
Hi @mna,

I have reorganized the code to address compatibility issues. This means we don't need to wait for the 1.23 release to merge (the merge timing is up to you).

If the user is using a version before 1.23, they can use the call function method. If the user is using version 1.23 or above, they can use the for-range method.

See the implementation in commit 9489580edc4074268e457e47b1418544abf655c6.

Thanks.

I believe the CI will pass this time 😂

mna:
Thanks for taking the time to do this, but I prefer to wait for the 1.23 release and not keep those conditional compiled files around - these tend to be useful for a short time as Go users typically upgrade quite fast to the latest versions. This will also allow enforcing the right toolchain in `go.mod`. Also, I think it's a bit nicer to use the `iter.Seq2` type and the actual for-range in the test, as that's how this is expected to be used.

But don't worry about it, I'll just revert the latest commit when I merge the PR!

mna:
I saw that Go 1.23 has been released, I'll try to get this merged later this week.

amikai:
Hi @mna 
You might have forgotten, so let me remind you because I want to use this feature 😂.

mna:
@amikai yes, thanks for the ping! I'll try to release this today!

PR #489, opened 2024-09-06
Merge preparation for amikai's itereach PR  

mna:
#483


andig:
Sorry if I'm nitpicking here. I'm not sure if user an `iter.Seq2` is the right thing to do here. Afaiaa the index part of the iteration is already provided by the `range` statement itself. The iterator here should be an `iter.Seq` if I'm not mistaken. Using `Seq2` just makes it confusing since that would usually be used with returning errors.

PR #545, opened 2026-05-23
perf: avoid per-node Selection allocation in winnow Not path

jvoisin:
The winnow() "Not" path routed through grep(), which called newSingleSelection() for every node, allocating a Selection and a 1-element []*html.Node slice on the heap per node, only to immediately unwrap it via s.Get(0) for m.Match().
This commit call m.Match(n) directly and pre-allocate the result slice.

It also adds BenchmarkNotMatcher to isolate the winnow path from compileMatcher overhead.

Here are my local benchmark results, for BenchmarkNotMatcher with 373 <li> nodes, and a pre-compiled matcher:

old ns/op   new ns/op   delta
22750       19350       ~-15%

old B/op    new B/op    delta
9384        3120        -66.7%

old allocs  new allocs  delta
11          2           -81.8%

### DIFF
--- a/bench_filter_test.go
+++ b/bench_filter_test.go
@@ -2,6 +2,8 @@ package goquery
 
 import (
 	"testing"
+
+	"github.com/andybalholm/cascadia"
 )
 
 func BenchmarkFilter(b *testing.B) {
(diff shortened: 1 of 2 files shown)


mna:
Thanks, again mentioning if you have interest in doing any more optimizations, there are a couple more calls to `grep` that immediately de-wrap the single selection with `s.Get(0)` that could benefit from that change.

PR #550, opened 2026-05-23
perf: skip dedup map in mapNodes for single-node selections

jvoisin:
mapNodes always allocated a map[*html.Node]bool for deduplication, even when called with a single source node where duplicates cannot arise. Short-circuit to return the callback result directly when len(nodes)==1.

This benefits all traversal methods (Find, Children, Parent, Next, Prev, etc.) when operating on a single-node selection, which is the common case after First(), Eq(), or directly on a Document.

name      old ns/op   new ns/op   delta
Find-8    23200       14200       -38.8%

name      old B/op    new B/op    delta
Find-8    4632        1496        -67.7%

name      old allocs  new allocs  delta
Find-8    27          13          -51.9%


mna:
LGTM but while we're adding this, let's handle len(nodes) == 0 too.

jvoisin:
Sure

PR #555, opened 2026-05-25
perf: iterate n.Attr by index in getAttributePtr and removeAttr

jvoisin:
`for i, a := range n.Attr` copies an entire html.Attribute value into `a` on every iteration, only to look at `a.Key`. Switching to `for i := range n.Attr` and indexing into `n.Attr[i].Key` removes the copy.

Both helpers sit on very hot paths:
- getAttributePtr: Attr, AttrOr, SetAttr, getClassesAndAttr (used in AddClass, RemoveClass, ToggleClass and HasClass)
- removeAttr: RemoveAttr, setClasses

While there are no changes in B/op or allocs/op since the copy was stack-only, it speed things up by around 15% on local benchmarks.

### DIFF
--- a/property.go
+++ b/property.go
@@ -210,8 +210,8 @@ func getAttributePtr(attrName string, n *html.Node) *html.Attribute {
 		return nil
 	}
 
-	for i, a := range n.Attr {
-		if a.Key == attrName {
+	for i := range n.Attr {
+		if n.Attr[i].Key == attrName {
 			return &n.Attr[i]
 		}
 	}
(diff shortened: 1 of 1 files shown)


mna:
> copies an entire html.Attribute value into a on every iteration

The `html.Attribute` is 3 strings: https://pkg.go.dev/golang.org/x/net/html#Attribute. This is an unncessary micro-optimization, I understand it can show a 15% "speed-up", but we're probably talking a few nanoseconds.

PR #556, opened 2026-05-26
perf: replace recursive closure in Text() with iterative subtree walk

jvoisin:
Text() built a self-referencing closure `var f func(*html.Node); f = ...` to recurse into descendants and accumulate text. The closure capture of its own variable forces the compiler to box `f`, and each recursive call pays a Go stack-frame setup.

This commit replaces it with an iterative depth-first walk using FirstChild/NextSibling/Parent pointers, anchored at each root in s.Nodes so the walk never escapes the original subtree.

Allocations are unchanged (the remaining 6/15/19 allocs come from strings.Builder grow-doubling), but ns/op drops by roughly 30-40% across shallow and deep subtrees:

name                old time/op  new time/op  delta
Text-8              1.86µs       1.11µs       -40.4%   (h2, n=14)
TextDeep-8          126µs        86µs         -31.6%   (whole body)
TextManyShallow-8   40.9µs       29.1µs       -28.9%   (all anchors)

All p=0.000 over 10 runs on arm64. B/op and allocs/op are bit-for-bit identical in all three benchmarks.

### DIFF
--- a/property.go
+++ b/property.go
@@ -60,21 +60,33 @@ func (s *Selection) SetAttr(attrName, val string) *Selection {
 func (s *Selection) Text() string {
 	var builder strings.Builder
 
-	// Slightly optimized vs calling Each: no single selection object created
-	var f func(*html.Node)
-	f = func(n *html.Node) {
-		if n.Type == html.TextNode {
+	// Iterative depth-first walk: Walks each root's subtree using FirstChild,
+	// NextSibling and Parent pointers, stopping when it ascends back to the
+	// root so it never escapes the original subtree.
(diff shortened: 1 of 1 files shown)


mna:
I'm not sold on that one, the resulting implementation is harder to read and has subtle conditions. I think just getting rid of the closure could be enough, and keep the recursive approach.

jvoisin:
I guess it's a matter of taste, I prefer the explicit iterative version with comments.

But alright, I'll try a version using an outlined function instead of a closure, benchmark it, and will send a pull-request if it's improving things.

jvoisin:
See #561

PR #563, opened 2026-05-27
perf: presize node-set maps in mapNodes, ClosestNodes and winnowNodes

jvoisin:
Three internal helpers build a map[*html.Node]bool from a known set of nodes (or use one as a dedup accumulator over a known number of input nodes) but allocate the map without a size hint. They then incur grow-and-double rehashing as entries are inserted.

Pass len(nodes) as the size hint at make time:

- traversal.go mapNodes: dedup map shared by appendWithoutDuplicates across every per-node result. mapNodes is the engine behind most traversal helpers (Parents, Children, Next/Prev, Siblings, Find*, Contents, ...), so the impact is broad.
- traversal.go ClosestNodes: target-node set built from nodes...
- filter.go winnowNodes: large-N path's lookup set built from nodes...

Double-digit performance gains across ~all benchmarks.

### DIFF
--- a/filter.go
+++ b/filter.go
@@ -149,7 +149,7 @@ func winnowNodes(sel *Selection, nodes []*html.Node, keep bool) []*html.Node {
 		})
 	}
 
-	set := make(map[*html.Node]bool)
+	set := make(map[*html.Node]bool, len(nodes))
 	for _, n := range nodes {
 		set[n] = true
 	}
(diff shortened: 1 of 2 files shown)


mna:
Thanks for the batch of PRs, love those small, focused improvements! Let me know if you have more coming in, once you're done with them I'll cut a new release.

jvoisin:
Thank you for your patience reviewing my late-night profiler-infused pull-requests :)

I'll keep sending them until there are no more low-hanging fruits, so likely at least until the end of next week

PR #569, opened 2026-05-30
perf: probe with MatchFirst in HasMatcher to avoid slice allocation

jvoisin:
HasMatcher only needs to know whether a child subtree contains a match, but it tested len(m.MatchAll(c)) > 0 after wrapping the matcher with SingleMatcher. SingleMatcher.MatchAll computes MatchFirst internally and then wraps the result in a throwaway []*html.Node{node}, allocating a one-element slice per matching child subtree just to check a boolean.

When the matcher exposes MatchFirst, probe it directly and test the returned node against nil, avoiding the allocation entirely. This is the same trick already used in SingleMatcher.MatchAll, just applied at the HasMatcher call site so the intermediate slice is never built in the first place. The SingleMatcher path is kept as a fallback for matchers that don't implement MatchFirst.

```
benchstat (count=10):

      | before       |              after   |
      | sec/op       |    sec/op     vs base|
Has-8 | 2.675µ ± 17% | 2.208µ ± 17%  -17.47%|

      | before       |             after    |
      |  B/op        |    B/op     vs base  |
Has-8 | 360.0 ± 0%   | 240.0 ± 0%  -33.33%  |

      | before       |             after    |
      |allocs/op     | allocs/op   vs base  |
Has-8 | 20.00 ± 0%   |  6.000 ± 0%  -70.00% |
```

### DIFF
--- a/filter.go
+++ b/filter.go
@@ -88,27 +88,15 @@ func (s *Selection) Has(selector string) *Selection {
 func (s *Selection) HasMatcher(m Matcher) *Selection {
 	result := make([]*html.Node, 0, len(s.Nodes))
 
-	// If the matcher can probe for the first match in a subtree without
-	// building a result slice, use that: it avoids a throwaway one-element
-	// slice allocation for every matching child subtree.
-	if mf, ok := m.(interface {
-		MatchFirst(*html.Node) *html.Node
-	}); ok {
-		for _, n := range s.Nodes {
-			for c := n.FirstChild; c != nil; c = c.NextSibling {
-				if c.Type == html.ElementNode && mf.MatchFirst(c) != nil {
-					result = append(result, n)
-					break
-				}
-			}
-		}
-		return pushStack(s, result)
-	}
-
-	m = SingleMatcher(m)
+	// Manually create a singleMatcher (rather than using SingleMatcher, which
+	// returns a Matcher interface) so we can call its MatchFirst method
+	// directly. MatchFirst probes for the first match in a subtree without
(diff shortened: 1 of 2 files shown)


filter.go
```diff
@@ -87,6 +87,24 @@ func (s *Selection) Has(selector string) *Selection {
...
+			for c := n.FirstChild; c != nil; c = c.NextSibling {
+				if c.Type == html.ElementNode && mf.MatchFirst(c) != nil {
+					result = append(result, n)
+					break
+				}
+			}
+		}
+		return pushStack(s, result)
+	}
+
```
mna: This performance improvement should be done a bit differently so it doesn't add complexity in this method, but abstracts it in `singleMatcher`, where it can be reused elsewhere as needed.

`singleMatcher` should implement a `MatchFirst` method, which calls the underlying Matcher's `MatchFirst` if it exists, otherwise falls back to `MatchAll` and return the single-item slice.

I.e. today's `singleMatcher.MatchAll` should become `singleMatcher.MatchFirst`, and `singleMatcher.MatchAll` should call `singleMatcher.MatchFirst` and wrap it in the single-item slice so that for external package users, `SingleMatcher()` returns a matcher that stops at the first one but wrapped in a slice (to respect the method signature), but internally we can avoid that slice.

The `singleMatcher` would need to be manually created here so we have access to the non-interface-method `MatchFirst` without type-casting.

Let me know if that isn't clear :)

jvoisin: I think™ I got your meaning, let me know if what I did is what you had in mind. Also, feel free to squash the two commits when merging them.

PR #570, opened 2026-05-30
perf: skip dedup map in getChildrenNodes

jvoisin:
getChildrenNodes went through mapNodes, which builds a map[*html.Node]bool to deduplicate results merged across source nodes. That dedup is pointless in this case: every child has exactly one parent, so the children of distinct (already-unique) source nodes form disjoint sets and can never collide. Walk the sources directly into the result instead, keeping a single-node fast path that returns the child slice without an extra copy.

This mirrors the disjoint-set reasoning already applied to the parent traversal helpers.

benchstat -count=10:

                    sec/op       B/op           allocs/op
Contents-8          -55.52%      -67.63%        -19.23%
ContentsFiltered-8  -15.25%      -33.61%        -9.68%
ChildrenFiltered-8  -32.84%      -3.23%         -7.69%
Children-8          ~ (single-node fast path unchanged)


mna:

PR #571, opened 2026-05-31
perf: presize result slice in getChildrenWithSiblingType

jvoisin:
The collect-all sibling traversals (Siblings, NextAll, PrevAll, Contents and their filtered variants) grew the result slice via append with no capacity hint, paying repeated grow-and-double reallocations.

This commit runs the existing iterator once to count the matching siblings, then presize the result slice before collecting. The Until cases are left alone, as their predicate would run twice, and the single-result Next/Prev cases need no allocation.

benchstat (count=10):

  allocs/op    Siblings -57%, SiblingsFiltered -54%, NextAll -60%,
               NextAllFiltered -56%, PrevAll -55%, PrevAllFiltered -53%,
               Contents -10%, ChildrenFiltered -17% (geomean -35%)
  B/op         geomean -6% (-9% to -11% on the All paths)
  sec/op       PrevAll -44%, Next -29%, Contents -27% (geomean -20%)


mna:
Not worth the early return, the 0 change will be fast enough anyway with an empty iteration next.

PR #572, opened 2026-05-31
perf: inline dedup loop in getParentNodes

jvoisin:
getParentNodes routed Parent/ParentFiltered through mapNodes with a closure returning a one-element []*html.Node{n.Parent} per source node. On a sibling selection nearly all of those slices point at the same parent and are immediately deduped away.

This commit walks s.Nodes directly instead: for a single node, return its parent with no map; for many, dedup parents into a shared set while appending. This removes the per-node slice allocation at the source rather than cleaning it up downstream.

benchstat (count=10, all p=0.000):

  allocs/op    Parent 384->7 (-98%), ParentFiltered 392->16 (-96%)
  sec/op       Parent -78%, ParentFiltered -59%
  B/op         both -22%


mna:

PR #577, opened 2026-06-01
perf: skip mapNodes dedup map in FindNodes

jvoisin:
Each source node passed to FindNodes maps to at most itself (when it is a descendant of the current selection), so there is nothing to deduplicate across distinct sources. Replace the mapNodes call and its per-node callback/slice allocations with a direct append loop.

Duplicate input nodes are still discarded with a cheap isInSlice check (run before the more expensive sliceContains tree walk) so the resulting Selection stays a proper set, matching the previous behaviour.

benchstat:
  sec/op  -43.68%
  B/op    -82.00%
  allocs    85 -> 9

Add TestFindNodesDuplicateInput to cover the duplicate-input case.


mna:

PR #578, opened 2026-06-03
perf: skip make+copy in cloneNode for attribute-less nodes

jvoisin:
Guard the Attr slice initialization with len(n.Attr) > 0. make([]T, 0) is zero-alloc in Go (returns a header into runtime.zerobase), so this does not change B/op or allocs/op, but it removes a runtime.makeslice call and a copy() call per attribute-less node. Text nodes (the majority of nodes in real HTML) and many element nodes have no attributes, so the saving applies to most of the recursion.

  BenchmarkCloneDocument       -5.7%  (844463 -> 796520 ns/op)
  BenchmarkSelectionClone      -7.2%  ( 68557 ->  63654 ns/op)
  BenchmarkSelectionCloneAll   -6.8%  (9459930 -> 8813318 ns/op)
  B/op, allocs/op:             unchanged

So nothing ground-breaking, but the change is minimal, so worth it.

### DIFF
--- a/manipulation.go
+++ b/manipulation.go
@@ -608,10 +608,14 @@ func cloneNode(n *html.Node) *html.Node {
 		Type:     n.Type,
 		DataAtom: n.DataAtom,
 		Data:     n.Data,
-		Attr:     make([]html.Attribute, len(n.Attr)),
 	}
 
-	copy(nn.Attr, n.Attr)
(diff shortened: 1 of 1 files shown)


mna:
Calling `make` with a len of 0 will not allocate a backing array, so it's a very cheap operation. As explained elsewhere, I won't merge those micro-optimizations.

PR #579, opened 2026-06-03
perf: hoist " class " wrap out of per-node loops in AddClass/RemoveClass

jvoisin:
The space-padded form of each input class was being rebuilt with `" " + c + " "` inside the per-node loop, allocating a fresh string per (node, class) pair. Pre-compute the wrapped form once per call into a small helper slice. In AddClass, the bare "class " append form is just `spaced[i][1:]` (a free re-slice), so no second helper is needed.

  BenchmarkAddClass     -38.2%  (90454 -> 55875 ns/op)
  BenchmarkRemoveClass  -41.2%  (81560 -> 47958 ns/op)

allocs/op: +4 per call (3 wrapped strings + 1 slice header) replacing the previous O(s * classes) per-node concat allocs.

### DIFF
--- a/property.go
+++ b/property.go
@@ -118,11 +118,18 @@ func (s *Selection) AddClass(class ...string) *Selection {
 	}
 
 	tcls := getClassesSlice(classStr)
+	// Pre-compute the " class " form once per input class, used to test
+	// for existence in the (already space-padded) current class string.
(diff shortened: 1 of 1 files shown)


mna:
Sorry, this is way too micro-optimization for the use-case of goquery, which typically deals with network and database requests.

PR #580, opened 2026-06-03
perf: detach children inline in Empty()

jvoisin:
Empty() called n.RemoveChild(c) in a loop where c was always the FirstChild. RemoveChild's general logic is overkill for that case: the panic-check on c.Parent is always true, the PrevSibling branch is always nil, the FirstChild reassignment is overwritten on every iteration, and the LastChild check only ever matches on the final child.

Walk the sibling chain once via NextSibling, nil out each detached child's parent/sibling pointers, and clear n.FirstChild / n.LastChild exactly once at the end.

### DIFF
--- a/manipulation.go
+++ b/manipulation.go
@@ -169,10 +169,20 @@ func (s *Selection) Empty() *Selection {
 	var nodes []*html.Node
 
 	for _, n := range s.Nodes {
-		for c := n.FirstChild; c != nil; c = n.FirstChild {
-			n.RemoveChild(c)
+		// Detach all children in one pass. Since each child is being
+		// removed from its only parent and ends up with no siblings,
+		// we can inline what RemoveChild does for the FirstChild case
+		// and skip the per-iter FirstChild/LastChild bookkeeping.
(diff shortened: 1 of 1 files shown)


mna:
So I have a growing _malaise_ with those perf PRs, in that they consistently and increasingly break a nice property of the codebase where there's a small number of core functions that are reused and composed to build a large API surface. A very common approach to those performance improvements are to inline some of that code reuse and logic in order to remove the parts that are unnecessary. Of course it results in more efficient runtime and less allocations, but very importantly, it adds complexity, code and logic duplication and makes the project harder to maintain (changing one thing here means remembering to tweak it there and there and ...).

I'm gonna pause merging those types of PRs for now, I want to think about them a bit more and may close them eventually if I'm not comfortable with the change. I appreciate the dedication to extract top performance, but it's important to put this package in its typical context, which is often in web-based and/or database-backed applications, where milliseconds and megabytes are usual metrics. The package has been very stable for a long time and I want to ensure it stays that way.

mna:

PR #581, opened 2026-06-03
perf: avoid temp html.Attribute in removeAttr's swap-and-shrink

jvoisin:
The tuple-form

    n.Attr[i], n.Attr[len-1], n.Attr =
        n.Attr[len-1], html.Attribute{}, n.Attr[:len-1]

forces the compiler to evaluate all three RHS expressions into temporaries before writing any LHS, materialising a full html.Attribute (three string headers, 6 words) on the stack on every call.

This commit splits it into sequential statements lets the compiler store n.Attr[last] directly into n.Attr[i], memclr the trailing slot in place, and reslice. While the performance gains are negligible (~3%), the main gain here is readability, as it's not super-duper-obvious what was going on here.

### DIFF
--- a/property.go
+++ b/property.go
@@ -263,8 +263,10 @@ func getClassesSlice(classes string) []string {
 func removeAttr(n *html.Node, attrName string) {
 	for i, a := range n.Attr {
 		if a.Key == attrName {
-			n.Attr[i], n.Attr[len(n.Attr)-1], n.Attr =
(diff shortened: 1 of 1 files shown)


mna:
Again, this is way too micro-optimized for goquery. As a rule of thumb, if the improvement requires working against Go idioms / changing the coding style to "game" the compiler, it's too low-level for this package.

PR #587, opened 2026-06-05
cleanup: factor getSiblingsNodes helper, rename for-Until variant

jvoisin:
Most callers of getSiblingNodes pass nil, nil for the trailing untilm/untilNodes args, the until criteria only apply to the Next/Prev *Until family. This commit splits the function in two:

1. Rename getSiblingNodes -> getSiblingNodesUntil; the name now matches the parameter shape.
2. getSiblingNodes(nodes, st) is now a helper that calls getSiblingNodesUntil(nodes, st, nil, nil).

It makes the code a bit more clear, as more than once I wondered what where the last two parameters from getSiblingNodes


mna:
I get the idea as it's a bit nicer at the call site but I don't think it's super clear why there's a get sibling nodes and a get sibling nodes until. Don't think it's worth the change.

PR #596, opened 2026-08-18
Add top-level Text function for BeautifulSoup-style text extraction

### LINKED ISSUE #443: How to achieve the effect of BeautifulSoup get_text？
How to achieve the effect of BeautifulSoup get_text？
```
def extract_content(html: str) -> str:
    soup = BeautifulSoup(html, "html.parser")
    text = soup.get_text(separator=" ", strip=True)
    return text

### PR DESCRIPTION
**Pull request by ChrisJr404**
This adds `goquery.Text(s *Selection, opts *TextOptions)`, a top-level counterpart to the `Selection.Text` method that gives control over how text is extracted, so callers can get clean, readable text out of a document without hand-rolling a node walk.

This is the feature discussed in #443, where the request was to offer something similar to Python BeautifulSoup's `get_text`. As suggested there, it is a package-level function rather than a `Selection` method (to keep the method set aligned with the jQuery API, like `NodeName` and `OuterHtml`), and it is "general enough" via a small options struct:

- `Separator` is inserted between the contents of consecutive text nodes.
- `Trim` strips leading/trailing whitespace from each text node and drops the ones that become empty (the insignificant whitespace from source indentation).
- `Keep` is an optional predicate called per text node; returning false excludes it, which is how a caller drops the text of `<script>`/`<style>` elements by inspecting the node's parent.

A `nil` (or zero-value) `TextOptions` makes `Text` behave exactly like `Selection.Text`, so the default is unsurprising and the traversal semantics match the existing method.

Example:

```go
text := goquery.Text(doc.Find("#content"), &goquery.TextOptions{
    Separator: " ",
    Trim:      true,
    Keep: func(n *html.Node) bool {
        if p := n.Parent; p != nil && p.Type == html.ElementNode {
            return p.Data != "script" && p.Data != "style"
        }
        return true
    },
})
```

Tests cover the nil-options equivalence to `Selection.Text`, the separator/trim behaviour, the `Keep` filter, and a multi-node selection; a runnable `ExampleText` is included as documentation. `go test ./...`, `go vet ./...`, and `gofmt` are all clean.

Closes #443.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/PuerkitoBio/goquery"
+	"golang.org/x/net/html"
 )
 
 // This example scrapes the reviews shown on the home page of metalsucks.net.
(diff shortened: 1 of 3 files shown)


mna:
Hello Chris,

Thanks for this PR! Took a quick glance this morning and it looks great, I will wait to merge for when I have a bit more time to dive into it. In the meantime, I noticed some edge cases that should probably be covered by a test, if you have a moment to update it - I'm thinking of calling `goquery.Text` with an empty selection (e.g. doc.Find that doesn't match anything) and one where every node is filtered-out by the Keep option.

Thanks again,
Martin

ChrisJr404:
Added the two cases you asked for, empty selection and everything filtered out by Keep. Thanks, Martin.

mna:
Looks great, thanks for the contribution!

PR #599, opened 2026-09-04
HasMatcher misses direct children that match the selector

udf2457:
Cascadia's MatchFirst(c) searches c's descendants, not c itself. So an element whose direct child matches the selector is never selected: doc.Find("div").Has("span") on <div><span>x</span></div> returns an empty selection. The MatchFirst "optimization" changed the semantics from "any descendant matches" to "any grandchild-or-deeper matches". Fix: also test the child directly — if m.Match(c) || sm.MatchFirst(c) != nil — or revert to the containment-based implementation (s.HasNodes(m.MatchAll(s.document.rootNode)...)).

### DIFF
--- a/filter.go
+++ b/filter.go
@@ -88,15 +88,15 @@ func (s *Selection) Has(selector string) *Selection {
 func (s *Selection) HasMatcher(m Matcher) *Selection {
 	result := make([]*html.Node, 0, len(s.Nodes))
 
-	// Manually create a singleMatcher (rather than using SingleMatcher, which
-	// returns a Matcher interface) so we can call its MatchFirst method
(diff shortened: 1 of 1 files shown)


mna:
Please provide a test that fails without the fix and passes with the fix.

mna:
AFAICT this is wrong:

> Cascadia's MatchFirst(c) searches c's descendants, not c itself.

See https://pkg.go.dev/github.com/andybalholm/cascadia#Selector.MatchFirst

> MatchFirst returns the first node that matches s, from n and its children.

And its implementation: https://github.com/andybalholm/cascadia/blob/v1.3.5/selector.go#L161

Issue #139, opened 2016-12-05
<noscript> causes selector to fail


nathan-osman:
Consider the following program:

```go
package main

import (
	"fmt"
	"strings"

	"github.com/PuerkitoBio/goquery"
)

const data = `<noscript><a href="http://example.org">click this link</a></noscript>`

func main() {
	d, err := goquery.NewDocumentFromReader(strings.NewReader(data))
	if err != nil {
		fmt.Println(err)
		return
	}
	a, ok := d.Find("noscript a").Attr("href")
	fmt.Printf("URL: '%s', %t\n", a, ok)
}
```

The expected output is:

    URL: 'http://example.org', true

But instead the output is:

    URL: '', false

Changing `noscript` to `div` in both the document and selector causes the expected output, so the problem seems to affect only `<noscript>` elements.

nathan-osman:
After further investigation, the problem appeared to originate in Cascadia:

```go
package main

import (
	"fmt"
	"strings"

	"github.com/andybalholm/cascadia"
	"golang.org/x/net/html"
)

const data = `<noscript><a href="http://example.org">click</a></noscript>`

func main() {
	n, err := html.Parse(strings.NewReader(data))
	if err != nil {
		fmt.Println(err)
		return
	}
	s, err := cascadia.Compile("noscript a")
	if err != nil {
		fmt.Println(err)
	}
	fmt.Println(len(s.MatchAll(n)))
}
```

Before I could file a bug there, however, I came across this: https://github.com/andybalholm/cascadia/issues/14

> "The net/html parser parses the document as if javascript were enabled. Because of that, the contents of noscript elements are just a single text node, not parsed HTML elements."

Now it looks like the bug exists in the `golang.org/x/net/html` package. Indeed, there is an open bug there for this very problem: https://github.com/golang/go/issues/16318

Sadly, it hasn't been fixed yet. :cry:

PuerkitoBio:
Hello Nathan,

Thanks for looking into this. Makes sense that this is at the html parser level, would be nice if it provided the option to set javascript on or off for parsing. I'll keep the issue open until some decision is made in the parser.

Martin

Arnold1:
just noticed the same issue :)

machinae:
For those looking for a workaround, re-parsing the content of the noscript tag seems to do the trick.
```go
s.Find("noscript").SetHtml(s.Find("noscript").Text())
```

Arnold1:
[...]

PR #602, opened 2026-09-05
Expose html parse options (addresses #139, noscript)

dualfroz:
## Problem

Refs #139.

## Background

#139 reports that no selector can reach inside `<noscript>`. As nathan-osman
found at the time, that is the html parser: it assumes scripting is enabled,
and per the HTML spec a `<noscript>` element is then parsed as raw text. Your
reply was:

> would be nice if it provided the option to set javascript on or off for
> parsing. I'll keep the issue open until some decision is made in the parser.

That decision has since been made upstream --
`golang.org/x/net/html` has `ParseWithOptions` and
`ParseOptionEnableScripting(bool)` (present in the v0.58.0 the module already
requires). goquery calls `html.Parse`, so there is currently no way for a
caller to reach it.

## Change

One additive constructor:

```go
func NewDocumentFromReaderWithOptions(r io.Reader, opts ...html.ParseOption) (*Document, error)
```

`NewDocumentFromReader` and every other exported symbol are untouched, so this
is source and binary compatible.

```go
doc, err := goquery.NewDocumentFromReaderWithOptions(r,
	html.ParseOptionEnableScripting(false))
href, _ := doc.Find("noscript a").Attr("href")   // "http://example.org"
```

## What this does and does not fix

Verified against `x/net v0.58.0`, `noscript a` on the same markup with scripting
on and off:

| document | scripting on | scripting off |
|---|---|---|
| `<body><noscript><a ...></noscript></body>` | 0 matches | 1 match |
| `<html><body><p>x</p><noscript><a ...></noscript></body></html>` | 0 | 1 |
| `<html><head><title>t</title></head><body><div><noscript><a ...></noscript></div></body></html>` | 0 | 1 |
| `<noscript><a ...></noscript>` (the issue's snippet, verbatim) | 0 | 0 |

So it fixes the case that matters in practice -- a `<noscript>` inside the body
of a real page -- but not the reporter's one-line snippet. In that snippet there
is no `<body>`, so the bare `<noscript>` is placed in `<head>`, and with
scripting off its content is real markup, which means the `<a>` is not allowed
there and the parser relocates it to `<body>`. The tree becomes
`<head><noscript></noscript></head><body><a ...></a></body>` and `noscript a`
correctly matches nothing. That is the spec's head/body handling for a
degenerate document rather than something goquery or the option can address,
and the anchor is reachable (`doc.Find("a")` finds it) -- it just is not a
descendant of the `noscript`.

I would rather say that plainly than claim #139 is closed by this. If you want
it closed anyway, the practical cases above are covered; if not, this at least
removes the reason it was blocked.

## Tests

- `TestNewDocumentFromReaderWithOptions` -- the same document parsed both ways,
  asserting no match with scripting enabled and the right `href` with it
  disabled.
- `TestNewDocumentFromReaderWithOptionsNoOptions` -- calling it with no options
  behaves like an ordinary parse.

Gate run on this branch:

- `go test ./...` -- ok
- `go test -race ./...` -- ok
- `go vet ./...` -- clean
- `gofmt -l type.go type_test.go` -- the only complaint in `type.go` is a
  pre-existing comment-indentation hunk that `gofmt -d` reports identically on
  pristine `master`; the added code is clean


mna:
Hello,

Thanks for diving into this. I'm a little hesitant to add new API surface (especially constructors) for this as there's already a simple approach with `goquery.NewDocumentFromNode` to create a doc from an already-parsed (with whatever options required) html tree.

The PR currently would not handle parsing the `AfterHtml`/`BeforeHtml`/etc. functions that add html strings to the document - those end up calling `html.ParseFragment` without any html options. I think it would make the most sense for those calls to also use the same options that were provided to create the document, but again there's already a simple approach for those and it's to parse with `html.ParseFragmentWithOptions` directly and use `AfterNodes`/`BeforeNodes`/etc. instead.

I'll give it some more thought, I can see how it would be convenient for goquery users.

Martin

dualfroz:
narrowed it down. no new api beyond the one constructor, and the options now also apply to `AfterHtml`/`BeforeHtml`.

mna:
Thanks! PR looks great, just on the fence about adding yet another constructor function, I'll think about it some more but I will likely merge it, it is convenient and I could see the html package adding more options in the future and it would be nice to have that support in goquery.

PR #608, opened 2026-09-18
Traverse all filtered siblings with a nil stop selection

jakezwang:
When the stop selection is nil, `NextMatcherUntilSelection` and `PrevMatcherUntilSelection` inspect only the immediate sibling. Their selector-string wrappers inherit this behavior, so later matching siblings are omitted.

Use the all-sibling matcher variants, matching `NextUntilSelection(nil)` and `PrevUntilSelection(nil)`. Regression tests cover both directions and both filter APIs, including result order and `End()`.

Validation: full tests on Go 1.26 and 1.27, with race detection on 1.27; `go vet ./...`. All four new API cases fail before the fix.


mna:
Hello,

Thanks for contributing! I understand the idea, and it is arguably a better choice to fallback to Next/PrevAll instead of immediate, but that's a behaviour change for a very corner case AFAICT? It's fairly easy not to pass `nil` as the stop selection if someone doesn't want the immediate sibling behaviour?

> Their selector-string wrappers inherit this behavior

What do you mean by that? Selector strings that don't match any node still result in a non-nil (but empty) selection.

I'm pretty strict against breaking changes at this point in the package's life, so unless there's a very strong point to be made that this fixes a bug, I'm afraid I won't merge that change. 

Thanks,
Martin

jakezwang:
This was found through independent edge-case testing, not a real application failure. By "selector wrapper" I meant the string-selector variants `NextFilteredUntilSelection` and `PrevFilteredUntilSelection`, which delegate to the Matcher variants; I did not mean an external wrapper or downstream project. The proposed behavior makes a nil stop selection traverse all matching siblings, but I recognize that changing this established behavior is a compatibility decision. I do not have production impact evidence to justify that tradeoff.

mna:
Gotcha, thanks for the update, I'll close it then.

PR #609, opened 2026-09-20
Preserve SVG and MathML namespaces when cloning nodes

agammann:
Cloning a selection or document loses the Namespace field on every node. SVG and MathML elements consequently become nodes in the HTML namespace in the cloned tree.

Copy Namespace alongside the other node fields. Add a regression through both Selection.Clone and CloneDocument, covering SVG and MathML roots and their children.

Validation on Windows with Go 1.27.1: both regression subtests fail before the fix; go test ./... and go vet ./... pass afterward.


mna:
Thanks for the contribution!

